Issue #3, opened 2015-08-20
Fix false positive in no-string-based-set-xxx- rules


HamletDRC:
There are 3 rules with false positives. 
The unit tests are currently passing, but should actually be failing if we'd fix the issue. 

See: 
* NoStringBasedSetImmediateTestInput.ts
* NoStringBasedSetIntervalTestInput.ts
* NoStringBasedSetTimeoutFailingTestInput-false-positive.ts

HamletDRC:
I asked about this in the tslint chatroom. Response from weswigham was: 

    The typechecker is not reliable within tslint, as tslint is run per-file and not per-project
    So typed defined elsewhere in the project are missing, resulting in missing types after inference
    So while you can get at the typechecker within tslint, it won't be able to reliably find what you want

Still investigating.

PR #46, opened 2015-10-13
Allow TemplateExpressions to pass the no-string-based-set-timeout rule

chrismbarr:
in my AngularJS app the `no-string-based-set-timeout` rule was bombing out on me with the following error

```
[13:12:22] [gulp-tslint] error layout/side-menu.directive.ts[125, 3]: An error occurred visiting a node.
Walker: NoStringParameterToFunctionCallWalker
Node: 
		$window.setTimeout(() => {
			// do stuff here
		}, 200)
TypeError: Cannot read property 'flags' of undefined
```
In Angular it's is discouraged to directly call `setTimeout` or anything on the `window` object since in a non-browser test environment you would not have a `window` object.  Instead an AngularJS version of it called `$window` is injected and normal `window` functions are called from this instead.

I'm not 100% certain that my proposed fixed will work in all cases, but it worked in mine.  I'm not totally sure what a `TemplateFunction` in TypeScript is anyway.

### DIFF
--- a/src/AstUtils.ts
+++ b/src/AstUtils.ts
@@ -76,9 +76,9 @@ module AstUtils {
             return true;
         }
 
-        if (expression.getFullText() === 'functionArg') {
-            //AstUtils.dumpTypeInfo(expression, this.languageServices, this.typeChecker);
-        }
+        //if (expression.getFullText() === 'functionArg') {
+        //    AstUtils.dumpTypeInfo(expression, this.languageServices, this.typeChecker);
+        //}
 
         return false; // by default the expression does not evaluate to a function
     }
--- a/src/AstUtils.ts
+++ b/src/AstUtils.ts
@@ -31,6 +31,9 @@ module AstUtils {
         if (expression.kind === ts.SyntaxKind.FunctionExpression) {
             return true; // function expressions are OK to pass
         }
+        if (expression.kind === ts.SyntaxKind.TemplateExpression) {
+            return true; // template expressions are OK to pass
+        }
         if (expression.kind === ts.SyntaxKind.Identifier) {
             let typeInfo : ts.DefinitionInfo[] = languageServices.getTypeDefinitionAtPosition('file.ts', expression.getStart());
             if (typeInfo != null && typeInfo[0] != null && typeInfo[0].kind === 'function') {


HamletDRC:
Thank you so much for taking the time to do this. I will take a look tonight. Any pieces of sample code you have could easily be added to the unit tests. Do you have any other examples? 

Please know that the "no-string-parameter-to-x" rules have an open defect: #3 

The TypeScript TypeChecker is unreliable for a variety of reasons. These rules will certain stop you from passing a string literal but often fail on unexpected errors like you see here ("An error occurred visiting...". I am still looking at how to fix this issue.

chrismbarr:
The only false positives I ran across in my project were calls to `$window.setTimeout` just like what I posted above.  So nothing else to really add here as far as code samples go, unless you'd like to see the above code in context which I could provide if needed.

I actually wasn't even aware that it was possible to pass string parameters to `setTimeout` and the like!  I think for now I'm just going to disable these 3 `no-string-based-*` rules since they aren't an issue for my team right now.

HamletDRC:
I am pretty sure this is another case of the typescript versioning problem. 

A template expression is like this: 

    var data = 'Joe Smith';
    $window.setTimeout(`<div>${data}</div>`, 200)

But imagine this usage: 
    var data = '<script>alert(1)</script>';
    $window.setTimeout(`<div>${data}</div>`, 200)

So I think template strings should not be allowed. But your usage is a lambda, so this ts.SyntaxKind.TemplateExpression must have the wrong number mapped to it.

chrismbarr:
Ah ok, `TemplateExpression` makes sense now, but I think you are correct in that it must be mapped incorrectly.  I ran a test that seems to confirm this.

I have this as a test file

```typescript
class TestClass {
	constructor(private $window: angular.IWindowService) {

		//Arrow Functions
		setTimeout(() => { }, 100);
		$window.setTimeout(() => { }, 100);
		
		//Standard Functions
		setTimeout(function () { }, 100);
		$window.setTimeout(function () { }, 100);
		
		//Strings
		const alertNum = 1;
		setTimeout("alert(" + alertNum + ")", 100);
		$window.setTimeout("alert(" + alertNum + ")", 100);
		
		//TS Template Strings
		setTimeout(`alert(${alertNum})`, 100);
		$window.setTimeout(`alert(${alertNum})`, 100);
	}
}
```

And then I added the following code [around here in AstUtils.ts](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/AstUtils.ts#L27-L28) (~ line 27)
``` typescript
console.log("KIND: ", expression.kind, ts.SyntaxKind[expression.kind]);
console.log(expression.getFullText());
```

Now running this rule against that test code above gives this this output

```
[13:58:43] Starting 'TSLint:Newer'...
KIND:  172 ArrowFunction
() => { }
KIND:  172 ArrowFunction
() => { }
KIND:  171 FunctionExpression
function () { }
KIND:  171 FunctionExpression
function () { }
KIND:  179 BinaryExpression
"alert(" + alertNum + ")"
KIND:  179 BinaryExpression
"alert(" + alertNum + ")"
KIND:  181 TemplateExpression
`alert(${alertNum})`
KIND:  181 TemplateExpression
`alert(${alertNum})`
[13:58:43] [gulp-tslint] error TEST.module.ts[14, 3]: An error occurred visiting a node.
Walker: NoStringParameterToFunctionCallWalker
Node: 
		setTimeout("alert(" + alertNum + ")", 100)
TypeError: Cannot read property 'members' of undefined
[13:58:43] [gulp-tslint] error TEST.module.ts[15, 3]: An error occurred visiting a node.
Walker: NoStringParameterToFunctionCallWalker
Node: 
		$window.setTimeout("alert(" + alertNum + ")", 100)
TypeError: Cannot read property 'members' of undefined
[13:58:43] [gulp-tslint] error TEST.module.ts[18, 3]: Forbidden setTimeout string parameter: `alert(${alertNum})`
[13:58:43] [gulp-tslint] error TEST.module.ts[19, 3]: Forbidden setTimeout string parameter: `alert(${alertNum})`
[13:58:44] Finished 'TSLint:Newer' after 901 ms
Process terminated with code 0.
```

So... now it seems like it's correctly identifying the `TemplateExpression`s?  Hmm... Looking into this further, you'll notice that an `ArrowFunction` comes back as mapped to enum number 172, but if you look at the TypescriptService.d.ts file, you'll notice that [`ArrowFunction` is mapped to 164](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/typings/typescriptServices.d.ts#L196), however [`TemplateExpression` is mapped to 172](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/typings/typescriptServices.d.ts#L204)!

So I think that explains my confusion.  I saw the number 172 come back while debugging and just put in the check for `TemplateExpression` - which is probably incorrect.  I am not sure why it's now blowing up on a the string params now instead of the arrow functions like it did before though...

HamletDRC:
Thanks for looking into this. 

The problem is definitely that we test and *compile* against TypeScript 1.5 but you're running against 1.6. I just pushed a fix for this so both 1.5 and 1.6 are supported. Can you pull from master and try again?

HamletDRC:
I just added your scenarios to the unit tests, made the tests pass, and pushed again.

chrismbarr:
I now have the latest changes, but Visual Studio is giving me the following error:
```
Type '{ Unknown: number; EndOfFileToken: number; SingleLineCommentTrivia: number; MultiLineCommentTrivi...' is not assignable to type 'SyntaxKind'.
  Object literal may only specify known properties, and 'Unknown' does not exist in type 'SyntaxKind'.
```

[in `SyntaxKind.ts` on line 42](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/SyntaxKind.ts#L42) and on [line 301](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/SyntaxKind.ts#L301).

I tried adding `Unknown: number;` to the `SyntaxKind` interface, but then it just complained about the next member in that object that didn't exist in the interface.

chrismbarr:
This seems to be working for me now.  My above test code now correctly produces the expected TSLint errors.

Even more so, Visual Studio will complain if you attempt to pass strings into `setTimeout` so, I don't think these rules might even ben needed any longer.
image

PR #52, opened 2015-10-20
Set & declare var on same line

chrismbarr:
Useful for Visual Studio projects with the "Allow implicit `any` type" disabled.  Otherwise Visual Studio complains with the error:
`Error	TS7005	Variable 'languageServiceHost' implicitly has an 'any' type.`


HamletDRC:
Awesome, thanks for taking the time to fix this!

PR #56, opened 2015-10-20
Fixing index param

chrismbarr:
It takes strings, not numbers.


chrismbarr:
Yep, Visual Studio reported it as specifying the wrong type.  Now with all the PR's I made that you've merged in, everything finally compiles in VS now.

PR #59, opened 2015-10-20
Fixed NoConstantConditionRuleTest

DickvdBrink:
Test didn't do what the description said and I noticed this was thee same tests as the one on line 50


HamletDRC:
Thanks for noticing this.

PR #60, opened 2015-10-21
Create dist patch

chrismbarr:
Is there any reason why the compiled JS files aren't provided in this repo?  I'm not sure if this is something you all want, but I feel like it would be super helpful to just be able to download these rules & drop them in my project without all the current hassle I've been having with getting Visual Studio to compile them.

I tried to put them in a `/dist` foler, but I noticed [that was specifically ignored](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/.gitignore#L9) so I didn't want to mess anything up. Rename this folder if needed.


HamletDRC:
All the releases are in the repo. 

Just clone the /releases branch to git it: https://github.com/Microsoft/tslint-microsoft-contrib/tree/releases

HamletDRC:
Does that solve your problem? It is not the latest... although I suppose we could publish the latest .js there always

chrismbarr:
Ah ok, I wasn't aware they were kept in a separate branch.  It seems a bit counter-intuitive to keep them there, and like you mentioned, it's not up to date... so no.

`<opinion>`
Right now I am able to compile these, but I feel like I shouldn't have to.  I just want to download them and use them in my project. If I find a bug to fix or a way to improve the rule functionality, then and only then should I need to download the source files and try to compile them.
`</opinion>`

DickvdBrink:
Personally I don't like having the `js` files (generated by TypeScript) on version control.. because if you do a build an add some fixes later (and don't rebuilt/test) the files might not match the TS files..
maybe it is an option to publish a nightly version of the rules on npm under `tslint-microsoft-contrib@next` like the TypeScript team does?

chrismbarr:
That would be good as well.  Basically just a way to download the actual rules and be able to drop them into their project with no build step.

HamletDRC:
I have never heard of publishing the compiler output into the version control repository, but maybe it is more common that I know of. I am going to publish the .js to the release branch for now, update the readme with download links, and talk with @danielmanesku and @brndkfr about making a @next version

PR #72, opened 2015-10-27
Add symbol to list of allowed typeof values

DickvdBrink:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

### DIFF
--- a/src/validTypeofRule.ts
+++ b/src/validTypeofRule.ts
@@ -8,7 +8,7 @@ import Utils = require('./utils/Utils');
 export class Rule extends Lint.Rules.AbstractRule {
     public static FAILURE_STRING = 'Invalid comparison in typeof. Did you mean ';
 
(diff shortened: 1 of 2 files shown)


HamletDRC:
Ah, I had no idea. Thanks!

HamletDRC:
Thank you so much for taking the time to do this.

DickvdBrink:
No problem, love the extra tslint-rules! :)

HamletDRC:
Cool. On our side, we are trying to take every frequent code review comment and turn it into a rule. We have over 40 TypeScript devs on our project now, so the leads are getting a little buried!  I'm also trying to take the eslint rule list and reimplement them all so the TS community has parity with ES. 

If you have any rules in mind then I'd enjoy hearing them. I have dev time down to about 30 minutes a rule, so a lot can be implemented quickly.

PR #80, opened 2015-11-17
Allow the tests to check the optional rule parameters

chrismbarr:
Now these can be tested as well!  All existing tests still pass, and the `no-http-string` tests file now has 2 new tests to ensure that the optional parameters for this rule work as expected.


chrismbarr:
sorry, idk why it wants to add those two "merge pull request" commits from my branch as well.  I'll make  PR without those commits.

Issue #75, opened 2015-10-30
new rule: avoid confusing array constructors


HamletDRC:
The following creates a violation because "[]" is simpler and preferred.

      new Array()    // violation: prefer []

The following creates a violation because [4, 5] is simpler and preferred: 

      new Array(4,5) // violation: prefer [4, 5]

The following is a violation because it confusingly creates [undefined, undefined, undefined, undefined]. There is almost never a reason to do this.  

       new Array(4)  // violation: prefer [undefined, undefined, undefined, defined] or simply []

HamletDRC:
we can add these checks to the existing prefer-array-literal rule

PR #81, opened 2015-11-17
Allow the tests to check the optional rule parameters

chrismbarr:
Now these can be tested as well! All existing tests still pass, and the no-http-string tests file now has 2 new tests to ensure that the optional parameters for this rule work as expected.

### DIFF
--- a/tests/ChaiVagueErrorsRuleTests.ts
+++ b/tests/ChaiVagueErrorsRuleTests.ts
@@ -4,35 +4,35 @@
 /* tslint:disable:quotemark */
 /* tslint:disable:no-multiline-string */
 
-import TestHelper = require('./TestHelper');
+import TestHelper = require('./utils/TestHelper');
 
 /**
  * Unit tests.
  */
-describe('chaiVagueErrorsRule', () : void => {
+describe('chaiVagueErrorsRule', (): void => {
 
-    var ruleName : string = 'chai-vague-errors';
+    var ruleName: string = 'chai-vague-errors';
 
-    it('should pass on xxx', () : void => {
-        var script : string = `
+    it('should pass on xxx', (): void => {
+        var script: string = `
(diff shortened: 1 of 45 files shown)


HamletDRC:
the failures are probably because you moved TestHelper and the relative paths of files changed. 

Can we pass these options without having to pass null when they are not needed?

HamletDRC:
Hi Chris, 

Thanks for the contribution! I fixed the issue and pushed. You have two commits on the HEAD of master (1st is the TestHelper changes, 2nd is the http-string test changes).

What I changed: 
* TestHelper is not moved to /utils (it screwed up paths)
* TestHelper API does not change unless you need to pass the ruleOptions (so all those tests do not need to pass null, it is instead the default value)

HamletDRC:
If you're looking for an easy rule to write... check this one out: https://github.com/Microsoft/tslint-microsoft-contrib/issues/75

HamletDRC:
And the reason we have not been committing to this repo recently is that we are still working on enabling all the 1.0 rules in our own project. Once all the rules are enabled then we'll return to writing new rules :)

chrismbarr:
Ah excellent, I like your solution better, that's much easier to understand!

I am currently working on implementing these rules as well as some custom rules I've written already for internal projects at my company.  Not totally sure if I'm "allowed" to share those rules or not though 
¯\\\_(ツ)_/¯

Several of my custom rules are extremely simple, like:
* disallowing files over x number of lines
* disallowing functions/methods over x number of lines
* a few things specific to AngularJS that help enforce [John Papa's Angular Styleguide](https://github.com/johnpapa/angular-styleguide)
  - only allowing files to have one of a few possible suffixes
  -  enforcing and class name rules based on that file name suffix
  - rules around `$inject` and registering things with angular, etc.
  - rules for angular directive property naming/prefixing

PR #92, opened 2016-01-15
my own typescript v1.7 support

reversi-fun:


HamletDRC:
Unfortunately, this does not compile. 

However, I still think you have a version issue. Here are the rules: 
**TSLint version 3.2.x users**: use project tslint-microsoft-contrib version 2.x
**TSLint version 3.1.x users**: Unsupported
**TSLint version 3.0.x users**: Unsupported
**TSLint version 2.x users**: use project tslint-microsoft-contrib version 1.x

PR #103, opened 2016-02-24
Implement max-func-body-length rule

gregolsky:
Rule `max-func-body-length` validates the number of non-empty lines of functions, arrow functions and methods.

@HamletDRC This is the fixed version of my function line counting rule. Please notice I set the limit option for this rule to 4574 for this repository. The reason I did so is it validates `mocha` fixtures and those are all wrapped in one huge `describe()` function callback. There are 2 ways we can solve this:

* leave it as is - this way people would need to setup 2 `tslint` configurations : one for regular code and the second one for tests (assuming they go with `mocha` or similar) or just disable the rule for every test file (using `/* tslint-disable ...`

* hack it so it won't validate functions passed as arguments to certain functions like `describe()` or `it()`, however if we do it this way and enable it by default users may get confused.

Thoughts?


HamletDRC:
Great work. I made some changes to support ignoring these blocks.

See my commit: #1b4708561e8ccecd9959ec70c1b2d22208af9425

HamletDRC:
https://github.com/Microsoft/tslint-microsoft-contrib/commit/1b4708561e8ccecd9959ec70c1b2d22208af9425

gregolsky:
That's cool! Thank you.

HamletDRC:
I updated the /releases branch so you can use the rule right away by pointing NPM to the git commit

gregolsky:
:+1:

chrismbarr:
Does this also look at the length of a `constructor` ?

gregolsky:
@chrismbarr Well, it does not yet, but I created this PR a moment ago https://github.com/Microsoft/tslint-microsoft-contrib/pull/104.

PR #104, opened 2016-03-01
fix: add constructor length validation for max-func-body-length rule

gregolsky:
Adds support for constructor body length validation for `max-func-body-length` rule.
Also tweaked error messages a little bit, so it says method for methods and function for functions:
```
$ tslint -r tslint-rules aaa.ts 
aaa.ts[6, 5]: Max constructor body length exceeded in class John - max: 5, actual: 7
aaa.ts[16, 1]: Max function body length exceeded in function magic() - max: 5, actual: 21
aaa.ts[22, 13]: Max arrow function body length exceeded - max: 5, actual: 7
```

It also displays a class name, when constructor is too long.
It uses the general length option or `ctor-body-length` option, if provided.


HamletDRC:
merged and pushed. Thanks!

HamletDRC:
To use this in your project *before* the next release, add this dependency: 

    git+https://github.com/Microsoft/tslint-microsoft-contrib.git#69bd7e841e6542dc0a439d6e4a8bf3636a4e4a16

PR #114, opened 2016-04-22
Cleanup TestHelper.ts

chrismbarr:
I did this cleanup in small commits to help explain what I did and why.  Mostly this was based on recommendations made by Visual Studio 2015 & ReSharper.

The commit `a5f283c` is the most important one, since it was previously always causing compile issue for me in VS.  This fixes it so that it still works like before, it compiles in VS, and it's a little easier to read.

### DIFF
--- a/tests/TestHelper.ts
+++ b/tests/TestHelper.ts
@@ -3,7 +3,6 @@
 /// <reference path="../typings/chai.d.ts" />
 
 import * as Lint from 'tslint/lib/lint';
-import Linter = require('tslint');
 import fs = require('fs');
 import chai = require('chai');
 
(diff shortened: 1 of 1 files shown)


HamletDRC:
I'll take a look this weekend, thanks!

HamletDRC:
Looks good. a few comments: 

* we don't care about having interfaces start with I, so I took that out
* each commit should have a passing "grunt all". If not, then squash the commits together until you do (you'll see that I squashed your tslint fixes push into the correct commit)

HamletDRC:
Also, for this one: 

>> Move functions below so that they don't call a function before it appears

Can this really happen?

HamletDRC:
I merged it in. Thanks!

chrismbarr:
@HamletDRC Alright good comments, thanks!

About the functions appearing before they are called, that was just a suggestion made to me by Visual Studio 2015 and/or ReSharper 9.  When opening files I want to get rid of the squiggly underlines, so mostly why I did this was to appease my IDE.  Technically calling a function before it exist could be a problem, but I think that's practically very unlikely.

In my day-to-day typescript coding we don't use a lot of modules with exported members, so I'm not too familiar with good patterns regarding this.Perhaps enforcing the order of exported vs. non-exported members would make for a good TSLint rule!  There's already the `member-ordering` rule that enforces the order of class members depending on visibility.

PR #117, opened 2016-04-26
Simplify debugging doc

ggirou:


README.md
```diff
@@ -132,18 +132,12 @@ Debug code
...
-    # on *nix machines:
-    node --debug-brk /usr/local/bin/grunt mochaTest
-    # on Windows machines:
-    node --debug-brk /c/Users/[your alias]/AppData/Roaming/npm/node_modules/grunt-cli/bin/grunt
+    node-debug grunt mochaTest
 
-Open in browser:
-
-    http://127.0.0.1:8080/?ws=127.0.0.1:8080&port=5858
+The `node-debug` command will load Node Inspector in your default browser (works in Chrome and Opera only).
```
danielmanesku: Does the previous info work in Firefox? I don't know, didn't try it myself.
But if it does, maybe we should just list this as option, alongside the previous info.

ggirou: I just tried again previous commands and it doesn't work either in Firefox.

HamletDRC:
thanks!

PR #127, opened 2016-05-11
prefer-consts and no-empty-line-after-opening-brace rules.

iclanton:


HamletDRC:
Here is the NPM documentation on main: 

> The main field is a module ID that is the primary entry point to your program. That is, if your package is named foo, and a user installs it, and then does require("foo"), then your main module's exports object will be returned.
> This should be a module ID relative to the root of your package folder.
> For most modules, it makes the most sense to have a main script and often not much else.
> 

I think that '.' and 'package.json' are both incorrect values. There is no published module here so the json attribute should be missing. 

Why do you need this change? It sounds to me (without knowing much) that the tool you use to integrate this package is broken.

iclanton:
We're trying to use [`require.resolve`](https://nodejs.org/api/globals.html#globals_require_resolve)`('tslint-microsoft-contrib')` to determine the directory path to this project, but it checks for a "main" property that points to a file that exists, or points to a directory that contains an "index.{js|json|node}" file. Without with "main" property, or with a "main" property that doesn't point to anything, [`require.resolve`](https://nodejs.org/api/globals.html#globals_require_resolve) doesn't work, and that's the cleanest way to determine the location of a NPM package that doesn't land in a deterministic location.

Perhaps in this case it would make sense to point to a JSON file that lists all of the rules in the package?

HamletDRC:
@iclanton I pushed the change to package.json

I am going to test the new rules on Microsoft Social Engagement before pushing those rules. 

Awesome work, thank you so much! 

I am going to rename this pull request and leave it open until the rules are merged

HamletDRC:
I finished merging the prefer-const rule. @bvancea you will like this rule :)

HamletDRC:
no-empty-line-after-opening-brace is merged as well. Closing PR

PR #132, opened 2016-05-16
Created no-jquery-raw-elements rule

JoshuaKGoldberg:
Stops people from creating jQuery elements with manually concatenated strings.
@sethobrien @adamtreineke

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ Rule Name   | Description | Since
 `no-increment-decrement`        | Avoid use of increment and decrement operators particularly as part of complicated expressions | 0.0.1
 `no-inner-html`                 | Do not write values to innerHTML, outerHTML, or set HTML using the JQuery html() function. Writing values to innerHTML can expose your website to XSS injection attacks. All strings must be escaped before being rendered to the page.| 2.0.4
 `no-invalid-regexp`             | Do not use invalid regular expression strings in the RegExp constructor. Similar to the [ESLint no-invalid-regexp](http://eslint.org/docs/rules/no-invalid-regexp.html) rule| 1.0
+`no-jquery-raw-elements`        | Do not create jQuery elements with attributes using string concatenation.
 `no-missing-visibility-modifiers` | Class members (both fields and methods) should have visibility modifiers specified. THe Principle of Least Visibility guides us to prefer private methods and fields when possible. If a developer forgets to add a modifier then TypeScript assumes the element should be public, which is the wrong default choice. | 1.0
 `no-multiline-string`           | Do not declare multiline strings | 0.0.1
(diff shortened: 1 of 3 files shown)


HamletDRC:
@JoshuaKGoldberg do you have a link to some documentation saying why this is a security risk?

JoshuaKGoldberg:
Not off the top of my head. The logic is that it's both dangerous (adding unsanitized user input) and error-prone (accidentally losing track of `'`/`"`s). Looking around now...

HamletDRC:
@JoshuaKGoldberg is your team using this rule today? 

Passing a string literal to $ is a bad way to create html elements... but it is secure because a string literal is not unsanitized user input. 

Using string concatenation from user input and variables and then passing that to $ is also a bad way to create html elements and is potentially a security risk... but this rule does not catch that scenario. And I think a rule that flags any variable being passed to $ will be too broad. 

This rule catches any call to $ that passes a string literal that looks like it has an attribute inside of it. Is that really want you want the rule to catch?

HamletDRC:
Do you prefer something like this approach? http://stackoverflow.com/a/9760343/28946

    $('<div/>', {
        html: $('<h1/>', {
            html: title
        }).after(
            $('<div/>', {
                'text': content,
                'class': 'content'
            })
        )
    }

JoshuaKGoldberg:
> is your team using this rule today?
No, but we're considering it.

The rule is intended to also catch behavior like:

```typescript
$("<div className='" + className + "' />");
$("<" + tagName + "/>")
```

HamletDRC:
@JoshuaKGoldberg I'm not sure what to do with this one. 

I agree that this code should produce a violation because it is string manipulation for creating elements: 

    $("<" + tagName + " />");
    "<div className='" + className + "' >"

But I don't think these should produce violations because it is a static string and not user input: 

    $("<input readonly >");
    $("<input readonly='true' >");

Here is the rule that I propose: 

- Look for calls to jquery
- Allow parameters to be String Literals - this is OK
- If the expression passed to jquery is a BinaryExpression using the '+' operator... then create a violation if any of the sub-expressions contains the '>' or '<' signs

HamletDRC:
Hi @JoshuaKGoldberg 
I pushed the code based on my above comments: 

https://github.com/Microsoft/tslint-microsoft-contrib/commit/73e04f7128223c3f44c1c78884d3016f60aa8bfd

Please check out the unit tests especially and let me know if you want this rule expanded.

JoshuaKGoldberg:
I'll take a look soon! Apologies for the long wait time

HamletDRC:
@JoshuaKGoldberg the only thing that is missing from the rule is the idea that attributes embedded inside of string literals should be avoided. I think our recommendation is that we should avoid doing this, even in a string literal. So I'll quickly add that check back in.

JoshuaKGoldberg:
Just happened to notice jQuery's site *does* have this notice.

http://api.jquery.com/append/, under "Additional Notes".

PR #133, opened 2016-05-16
Added no-var-self rule

JoshuaKGoldberg:
This is a common practice in JavaScript projects converted to TypeScript.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ Rule Name   | Description | Since
 `no-unnecessary-semicolons`     | Remove unnecessary semicolons | 0.0.1
 `no-unused-imports`             | Remove unused imports | 0.0.1
 `no-with-statement`             | Do not use with statements. Assign the item to a new variable instead | 0.0.1
+`no-var-self`                   | Do not use `var self = this`; instead, manage scope with `() => {}` closures. This is a common practice in JavaScript projects converted to TypeScript.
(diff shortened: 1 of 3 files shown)


JoshuaKGoldberg:
Not sure why the build is failing; it passes locally and I did run the `grunt-create-rule` command. Same with #132.

HamletDRC:
Thanks, I will try to take a look later today! 

The no-function-expression and no-unnecessary-bind rules have a similar intent: use a lambda so you do not need to bind 'this'. 

I am interested to see what you do around nested lambdas where the inner lambda will need access to some self variable.

JoshuaKGoldberg:
Renaming to something other than `self`.

In fact, I'll change this later today to require the user to name which names are banned.

HamletDRC:
Should the rule just ban assigning 'this': 

    var x = this; // should be violation
    let self = this; // should be violation

I assume it should fail for any BinaryExpression where the operator is Equals and the right-hand-side is 'this'.

HamletDRC:
you should also add test cases for using let and const as well, and not just var.

JoshuaKGoldberg:
There are legitimate uses of assigning this. I'd rather not ban it altogether, though allowing "*"/-style wildcards for names can put that functionality in.

HamletDRC:
I suggest you do this: 
* Look for BinaryExpressions assigning the 'this' reference in a variable/let/const statement
* Add an option to the rule for "variableName" that is a regex
* default the regex to '.*'
* compare the variable/let/const on the right hand side to your regex and create a violation if it does matches.

HamletDRC:
This one is merged. Thanks @JoshuaKGoldberg

JoshuaKGoldberg:
Oh sorry I was actually intending on implementing your suggestions! I'll take a look at what isn't in and send a PR soon. Thanks!

JoshuaKGoldberg:
Looks perfect. :)

PR #156, opened 2016-06-07
Allowed simple open tags in no-jquery-raw-elements

### LINKED ISSUE #155: Simple unclosed tags are flagged by no-jquery-raw-elements
`TSLint failure: Replace complex HTML strings with jQuery API: $('<div>').`

The full line is:

```typescript
html: "<pre>" + $("<div>").text(message).html() + "</pre>",
```

...so it's important to know which part is being complained about.

JoshuaKGoldberg:
`$("<div>")` was previously being flagged as a violation.

Fixes #155.


JoshuaKGoldberg:
Looks like this was fixed in 367f5ffac71cbcf9a156b39949ee8076b0e23b17.

PR #159, opened 2016-06-22
More helpful failure message for noob developers

chrismbarr:
I've run into a situation at work where a less-experienced developer was trying to use a `for in` statement to loop over a standard array of strings, like so:

```typescript
const arr = ["foo", "bar", "baz"];
for (item in arr) {
  //do stuff
}
```

Then TSLint would produce the message:
> Do not use for in statements, use Object.keys instead: for (item in arr)

So the developer said "OK" and then went and changed the code to this:
```typescript
const arr = ["foo", "bar", "baz"];
Object.keys(arr).forEach((key: number) => {
  //do stuff
});
```

:neutral_face: 

Sooo... *technically* both ways actually do work for looping over flat arrays, but obviously they are not ideal.  I wanted to make the failure message for this rule a bit more explicit to better describe the available options.  Now the rule will produce this failure message:

>Do not use the 'for in' statement: 'for (item in arr)'. If this is an object, use 'Object.keys' instead. If this is an array use a standard 'for' loop instead.

### DIFF
--- a/src/noForInRule.ts
+++ b/src/noForInRule.ts
@@ -7,7 +7,10 @@ import {ErrorTolerantWalker} from './utils/ErrorTolerantWalker';
  * Implementation of the no-for-in rule.
  */
 export class Rule extends Lint.Rules.AbstractRule {
-    public static FAILURE_STRING = 'Do not use for in statements, use Object.keys instead: ';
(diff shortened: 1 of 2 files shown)


HamletDRC:
Great improvement. Feel free to do this as often as you can. Error messages are important.

PR #160, opened 2016-06-22
Trim whitespace from message in `no-multiline-strings`

chrismbarr:
on master, I keep running into this test failure with the `no-multiline-strings` rule when I run the tests locally.  It happened a long time ago to me and I ignored it as a fluke, but it's still happening to me today.

image

Basically it seems to expect the line break inside the reported string for the error message.  This was an easy thing to fix though, and now all tests pass for me with this small change.


HamletDRC:
I think the issue is your git line ending settings. The instructions in the readme.md say to change your git line endings globally before you check out. But this, of course, effects *all* your cloned git repos so it is not ideal. Anyway, merged and pushed.

PR #163, opened 2016-06-28
Fix trailing comma rule

Invizory:
`recommended_ruleset.js` contains the following rule:

```js
"trailing-comma": true,
```

This is actually a no-op because you should set `multiline` and `singleline` options in order to enable corresponding checks (see [trailing-comma rule doc](http://palantir.github.io/tslint/rules/trailing-comma/) and [recommended.ts](https://github.com/palantir/tslint/blob/master/src/configs/recommended.ts#L91)):

```js
"trailing-comma": [true, {"singleline": "never", "multiline": "always"}],
```

The rationale behind enforcing trailing commas for multiline statements is covered by [this article](https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8).

Hovewer, the established code style in `tslint-microsoft-contrib` assumes that trailing commas are disallowed so I've decided to set `"multiline": "never"` in `tslint.json` as a compromise and fix one redundant trailing comma.

### DIFF
--- a/recommended_ruleset.js
+++ b/recommended_ruleset.js
@@ -128,7 +128,7 @@ module.exports = {
         "one-line": [true, "check-open-brace", "check-catch", "check-else", "check-whitespace"],
         "quotemark": [true, "single"],
         "semicolon": [true, "always"],
-        "trailing-comma": true,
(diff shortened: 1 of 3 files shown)


HamletDRC:
merged and pushed, thanks!

PR #170, opened 2016-07-06
Added option to ignore comments in function body length

nebez:


HamletDRC:
I'll try to get to this later today... but I would not recommend setting this option. The problem with long functions is scroll-blindness: having to scroll the file in order to see a single function makes the function difficult to understand. Enabling this option would mean you have scroll blindness yet the tslint rule still passes. But I am happy to merge this if it's what you really want!

nebez:
Hey @HamletDRC. I'm inclined to agree with you now that I understand the
reasoning behind it.

I mostly worked on this PR for a specific use-case of mine where my
`max-func-body-length` rule was set to a very low number but some methods
had very long function bodies due to multi-line comments that detailed the
API consumption of HTTP routes.

Example:
```javascript
class ProductsController extends Controller {
    initializeRoutes(server: Server) {
        /**
         * @swagger
         * path: /products/:id
         * operations:
         *   -  httpMethod: GET
         *      summary: View a product
         *      notes: Returns a product based on numeric ID
         *      responseClass: Product
         *      nickname: product_view
         *      parameters:
         *        - name: id
         *          description: Product ID
         *          paramType: query
         *          required: true
         *          dataType: number
         */
        server.get('/products/:id', this.read);

        // ...
        server.post('/products', this.create);

        // ...
        server.delete('/products/:id', this.delete);
    }

    read(id: number): Response {
        // ...
    }

    create(attributes: any): Response {
        // ...
    }

    delete(id: number): Response {
        // ...
    }
}
```

Like I said, it's a very specific use-case and only cares for multi-line
comments. I could modify the linter rule to look for patterns like this
(swagger docs, js docs within functions, etc) or only care for multi-line
comments. Otherwise, in this case I could rename all of my route
initialization functions to share a common name (or prefix) and use the
regex option to ignore them similar to the mocha `describe` function.

I'm not sure that this is a common issue people are facing. Feel free to
close the PR if you think having this rule would be counter-intuitive.

HamletDRC:
Ah swagger :)

I'll make the commit today, I promise.

HamletDRC:
Really nice job on the code and tests. Thanks for the high quality submission!

Issue #171, opened 2016-07-06
new rule: no-unsupported-browser-code


AdamTReineke:
It's easy to accumulate lots of browser specific code when adding workarounds for browser bugs. This new rule would use the tslint.json to specify which browser versions you care about supporting, then specifically formatted comments to annotate all browser hacks. As you drop support for older browsers and change the tslint.json configuration, tslint would flag the comments so you can remove the code.

Some examples from our code base, annotated with new "Browser Specific" comments:

````typescript
// Called when we get a pointer down event for IE11
// Browser Specific: IE 11
document.body.addEventListener("pointerdown", function (event: MSPointerEvent) {
    if (event.pointerType !== "mouse") { ... }
}, false);

// Called when we get a pointer down event for IE10
// Browser Specific: IE 10
document.body.addEventListener("MSPointerDown", function (event: MSPointerEvent) {
    if (event.pointerType === event.MSPOINTER_TYPE_PEN) { ... }
    else if (event.pointerType === event.MSPOINTER_TYPE_TOUCH) { ... }
}, false);
````

````typescript
// Work around bug where this doesn't render on iOS by forcing the hardware to do it
// Browser Specific: Mobile Safari
if (Utilities.browser.isSafari && !Utilities.browser.isChrome && Utilities.browser.isMobile) {
	style["transform"] = "translate3d(0,0,0)";
}
````

If we dropped support for IE 10, the event listener would be flagged and we could remove that code from our code base.

Some additional thinking is needed to come up with a decent grammar for designating the browser versions. `@browserspecific <string>` (for JSdoc usages) or `Browser Specific: <string>` seem like good tags around the string itself.

A grammar something like `[Mobile] (Safari|Chrome|Firefox|FF|Internet Explorer|IE|Edge) [[>|<]<number>]` would lead to comments like `Mobile Safari < 8` or `Chrome > 45` or `IE`, which seems ok.

HamletDRC:
Great. idea. It seems pretty easy to implement. 

to configure the rule you could list your supported browers and have something like: 

    no-unsupported-browser-code: [ true, [
        "IE 10",
        "IE 11",
        "Chrome [4-9][0-9]",   // regex for 40 to 99
    ]]

HamletDRC:
Tagging feature request and microsoft-internal so we can make this a high priority. PRs welcome :)

PR #175, opened 2016-07-11
[Issue #171] New rule: no-unsupported-browser-code

nebez:
This implements feature-request #171

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Rule Name   | Description | Since
 `no-unnecessary-local-variable` | Do not declare a variable only to return it from the function on the next line. It is always less code to simply return the expression that initializes the variable. | 2.0.4
 `no-unnecessary-override`       | Do not write a method that only calls super() on the parent method with the same arguments. You can safely remove methods like this and Javascript will correctly dispatch the method to the parent object. | 2.0.4
 `no-unnecessary-semicolons`     | Remove unnecessary semicolons | 0.0.1
+`no-unsupported-browser-code`   | Avoid writing browser-specific code for unsupported browser versions. Browser versions are specified in the rule configuration options, eg: `[true, [ "IE 11", "Firefox > 40", "Chrome >= 45" ] ]`. Browser-specific blocks of code can then be designated with a single-line comment, like so: `// Browser specific: IE 10`, or with a jsdoc like this: `@browserspecific chrome 40`. | 2.0.10
 `no-unused-imports`             | Deprecated - This rule is now covered by TSLint's no-unused-variables rule. However, it can still be useful to enable this rule and pair it with the fix-no-unused-imports formatter. | 0.0.1
 `no-with-statement`             | Do not use with statements. Assign the item to a new variable instead | 0.0.1
 `no-var-self`                   | Do not use `var self = this`; instead, manage scope with arrow functions/lambdas. Self variables are a common practice in JavaScript but can be avoided in TypeScript. By default the rule bans any assignments of the `this` reference. If you want to enforce a naming convention or allow some usages then configure the rule with a regex. By default the rule is configured with `(?!)` which matches nothing. You can pass `^self$` to allow variables named self or pass `^(?!self$)` to allow anything other than self, for example| 2.0.8
@@ -133,7 +134,7 @@ Rule Name   | Description | Since
 Supported Formatters
 -----
 
-These formatters assume that you use the UTF-8 file encoding. They may not work if you have a different encoding. We are trying to fix this. 
+These formatters assume that you use the UTF-8 file encoding. They may not work if you have a different encoding. We are trying to fix this.
 
 Formatter Name          | Description | Since
 :----------             | :------------ | -------------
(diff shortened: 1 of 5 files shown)


HamletDRC:
@AdamTReineke please check out the unit tests submitted with this "no-unsupported-browser-code" rule. Does this meet your needs?

README.md
```diff
@@ -115,6 +115,7 @@ Rule Name   | Description | Since
 `no-unnecessary-local-variable` | Do not declare a variable only to return it from the function on the next line. It is always less code to simply return the expression that initializes the variable. | 2.0.4
 `no-unnecessary-override`       | Do not write a method that only calls super() on the parent method with the same arguments. You can safely remove methods like this and Javascript will correctly dispatch the method to the parent object. | 2.0.4
 `no-unnecessary-semicolons`     | Remove unnecessary semicolons | 0.0.1
+`no-unsupported-browser-code`   | Avoid writing browser-specific code for unsupported browser versions. Browser versions are specified in the rule configuration options, eg: \[true, \[ "IE 11", "Firefox > 40", "Chrome >= 45" \] \]. Browser-specific blocks of code can then be designated with a single-line comment, like so: `// Browser specific: IE 10`, or with a jsdoc like this: `@browserspecific chrome 40`. As you drop support for browsers, you can modify your supported browsers in the config to have the linter throw errors | 2.0.9
```
JoshuaKGoldberg: Nit: might look prettier if you remove the last sentence and wrapped the first code sample with code \`s too.

AdamTReineke:
Looks awesome!

nebez:
@AdamTReineke updated the README. Is that closer to what you had anticipated?

JoshuaKGoldberg:
@nebez looks good!

HamletDRC:
The reason I haven't pushed this yet is because of the scanner code. I think it is possible to just use a visitNode() override and test for the correct syntax kind on the node instead of creating your own scanner. The problem with the scanner is that you hard code the language variant and script target. There are probably problems with forcing .tsx files to the wrong language variant.

HamletDRC:
I just noticed you have the same scanner code in the function length rule.

nebez:
I see what you mean.
This was my first attempt at working with a linter (+ the func-length rule), I don't doubt that this isn't properly implemented. I misguidedly used existing tslint rules ([comment-format](https://github.com/palantir/tslint/blob/master/src/rules/commentFormatRule.ts), [no-trailing-whitespace](https://github.com/palantir/tslint/blob/2f3a20d90fd90e34a7fb31bcaed7b586b4a8039d/src/rules/noTrailingWhitespaceRule.ts), [jsdoc-format](https://github.com/palantir/tslint/blob/2f3a20d90fd90e34a7fb31bcaed7b586b4a8039d/src/rules/jsdocFormatRule.ts), etc.) to help point me in the right direction for the proper approach to working with the existence of comments in linter rules instead of finding the more appropriate approach with `visitNodeX` overrides.
I'll take some time today to re-visit my approach to both this rule and `max-func-length`. If I don't come up with anything, we can close this PR.
In the meantime anybody else could take a stab at implementing this rule and/or making the same changes with my last PR #170.

nebez:
Seems like all of the walkers for the rules mentioned earlier extend from [SkippableTokenAwareRuleWalker](https://github.com/palantir/tslint/blob/24017dc49c64d45c5652798e357997d7faab23fa/src/language/walker/skippableTokenAwareRuleWalker.ts) to specifically avoid walking through template nodes.

HamletDRC:
@nebez I looked at the tslint rule jsdocFormatRule. 
https://github.com/palantir/tslint/blob/master/src/rules/jsdocFormatRule.ts

It looks like they are using their own scanner as well and hard-coding a .ts file (not a .tsx file) type. 

Perhaps your code is the correct way to implement this. SHould we reopen this and do some more testing on .tsx files?

nebez:
@HamletDRC I have no experience with React. Do you have any ideas on testing .tsx files?

Also, like I mentioned in a previous comment, all the tslint linter rules that create their own scanner seem to extend from [skippableTokenAwareRuleWalker](https://github.com/palantir/tslint/blob/24017dc49c64d45c5652798e357997d7faab23fa/src/language/walker/skippableTokenAwareRuleWalker.ts), whereas the tslint-microsoft-contrib linter rules extend from [ErrorTolerantWalker](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/utils/ErrorTolerantWalker.ts). I haven't found a case where the ErrorTolerantWalker trips up on code yet, but then again my tests have been limited to non-.tsx only so I can't be sure there's a real difference.

HamletDRC:
As you see, we had to enable the "cla agreement bot" :)

The ErrorTolerantWalker is just a safety net so that an error in a rule does not fail someone's build. It is often the case that the type-system related rules throw error because it is not really possible to use the type checker from tslint. Anyway, it is not needed to subclass this. 

I added a method called AstUtils.getLanguageVariant that will give you the correct language variant. So we do not need to hardcode .js. I only wish I knew how to determine the compiler level (es5 or es6) instead of hard coding it.

nebez:
Thanks for the explanation. I just looked into `getLanguageVariant`, looks good!
I'll give this another go over the weekend and we can re-examine the implementation once it's complete.

nebez:
@HamletDRC I've updated the PR. Thoughts?

AdamTReineke:
Thanks @nebez and @HamletDRC!

PR #176, opened 2016-07-11
[Issue #170] Fixed wrong usage of SyntaxKind

nebez:
I wrongly used `ts.SyntaxKind` in my last PR. This should fix that.
Apologies!


HamletDRC:
merged and pushed

PR #178, opened 2016-07-13
Maintain spacing when changing let to const

chrismbarr:
Previously this would sometimes result in more space then were there originally like `const__= 4` (where `_` is a space)

### DIFF
--- a/src/fixPreferConstFormatter.ts
+++ b/src/fixPreferConstFormatter.ts
@@ -17,7 +17,7 @@ export class Formatter extends BaseFormatter {
             const start: number = failure.getStartPosition().getPosition();
 
             let leftSide: string = fileContents.substring(0, start);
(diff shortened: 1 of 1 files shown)


HamletDRC:
can you post some example input that would cause this feailure? 

I know this might fix the issue, but I think the problem is not the regex but instead how we are calculating what to replace. This seems related to the other "letr" issue

chrismbarr:
it seems adds a space regardless of how many spaces were there before.  This change finds `let` + however many whitespace chars are there and maintains them instead of always adding a hard-coded space

I think the reason this happened was because the `leftSide` var sometimes didn't always return the code with the whitespace intact. 
Now:
`let` :arrow_right: `const`
`let_` :arrow_right: `const_`
`let__` :arrow_right: `const__`
etc.

HamletDRC:
You actually need to put the * character inside the parens for this to work.

PR #237, opened 2016-09-09
Support for context.only in MochaAvoidOnlyRuleWalker

Przemek-at-Ais:
In Mocha .only can be added not only to describe and it, but also to context.


Przemek-at-Ais:
What else should be also adjusted so MochaAvoidOnlyRule would catch context.only usages? Thanks in advance for hints.

HamletDRC:
The issue is inside MochaUtils.isMochaTest. It defined a mocha test as something that had a top level describe. I updated it to include context. 

Great work! Thank you for submitting the PR, even with the failing unit test. This really helps us move forward even if it isn't 100% perfect on the first try.

Issue #64, opened 2015-10-22
Update development section of the doc


loicraux:
1) The Git configuration should be local, not global
2) Cloning with https rather than ssh, it requires no setup ?

HamletDRC:
loic, are you sure that local will work? You need to have this line endings set when you clone the repo. So there is no local config at that point. I think only global will work in this case.

loicraux:
First, I actually did not follow the steps for development, since I did not want to override my default global git settings. I think we should really remove these `git config --global` lines for the `README.md`. This is the original purpose of this ticket. But these line endings issues are always fun, so I did some further research in my train :)

So what actually happened to me:

- Since my global git setting is set to `autocrlf=true`, after cloning, all my text files in my working directory consequently had CRLF characters (that is `"r\n"`) as line endings, in particular `test-data\NoMultilineStringTestInput.ts`
- This particular rule `noMultilineStringRule.ts` being tested trims the faulty node text [this way](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/noMultilineStringRule.ts#L21) :
`let firstLine : string = fullText.substring(0, fullText.indexOf('\n'));`
- Eventually `chai.assert.deepEqual(actual, expected);` in `TestsHelper.ts` fails because the expected string do not have the `\r` char whereas the actual one has it :)

To fix the Development section of the `README.md` and get rid of the `git config --global` commands, one easy way would be to use this command :

	git clone git@github.com:... --config core.eol=lf --config core.autocrlf=input

This sets locally these settings before fetching of the remote changes are done. With this command, all files are checked out with LF line endings only, and Git ensures that any potential CRLF line endings will be normalized to LF on checkin to the DB.

Fine. But this is a bit unusual to pass these config settings before cloning or to the clone command itself. This is the old system of Git to handle line endings. We have this new system with the `.gitattributes` file that overrides any global or local git configuration and forces everybody to use the same git line endings settings, so we should be able to get rid of the `--config` options commands completely.

`* text=auto` in our current `.gitattributes` file set all files with any potential CRLF line endings will be normalized to LF on checkin to the DB, if those files are determined by Git to be text and not binary. This relies on Git's built in binary detection heuristics. But files would be checkout out with CRLF for Windows clients... Not what we want since the unit test would fail.

[...]

PR #248, opened 2016-09-15
fixes issue #64 by updating the dev doc

loicraux:
Just a small update in the doc for contributing to the repo so that git config is not global


HamletDRC:
thanks loic!

PR #263, opened 2016-09-20
Will allow us to use `extends` in our tslint.json

gaurav21r:
This will allow us to use 

```js
"extends": "tslint-microsoft-contrib"
```
in our project's `tslint.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "bugs": {
     "url": "https://github.com/Microsoft/tslint-microsoft-contrib/issues"
   },
-  "main": "package.json",
+  "main": "tslint.json",
   "keywords": [
     "tslint",
     "microsoft",


gaurav21r:
This is a very small change and https://github.com/rcorp/standard-project-structure/issues/15 is blocked because of it. Request for feedback / merge quickly ? :)

HamletDRC:
Ah, I always wondered how this worked. I can merge this in the next few hours. 

Do you need a release? Or can you configure your npm dependency like this: 

    "tslint-microsoft-contrib": "git@github.com:Microsoft/tslint-microsoft-contrib.git@releases"

We are making a release over the next day or two regardless.

HamletDRC:
If you ever have ideas for new rules then please don't hesitate to file an issue here!

gaurav21r:
@HamletDRC Thanks a lot for your quick response. Well a release would be great (even minor version number) but we can do without one as well. Incase that doesn't fit with your release strategy, It'll be great, if by the time we go live with https://github.com/rcorp/standard-project-structure (Maybe October 2nd week) we can get a release!

HamletDRC:
@iclanton This pull request changes main to tslint.json instead of package.json. Your last commit changed it to package.json with the comment "Pointing main in package.json to a real path." #8ab3195

If we change this to tslint.json then will it break something on your side? tslint.json is a real file in our npm package, so I think you will be OK:

gaurav21r:
@HamletDRC Thanks for the merge :smiley: ! I didn't understand 
> This pull request changes main to tslint.json instead of package.json. Your last commit changed it to package.json with the comment "Pointing main in package.json to a real path." #8ab3195

[I  changed `package.json` in this PR](https://github.com/Microsoft/tslint-microsoft-contrib/pull/263/commits/978b5f894a3d242ff4d9f1bba23044247325f0df). Are you saying we need to change `tslint.json`? But in the final commit you seemed to have changed `package.json` only? I'd love to learn the use case here for future ref.

HamletDRC:
@gaurav21r I think it was done correctly... but let me clarify my comment: 

* We changed package.json
* within that file we previously had "main": "package.json"
* now that file has: "main": "tslint.json"

I think it is all correct. I merged your PR as-is.

gaurav21r:
@HamletDRC Oh ok now I got it! Thanks for the quick release! Much appreciated :smile: 

Btw, it seems the latest release hasn't been updated on **npm**. Requesting you to please see that as well? Thanks!

HamletDRC:
@gaurav21r the npm package is there now. Our npm admin is in California so sometimes we have to wait a few hours :)

PR #282, opened 2016-09-28
fix aria-role-supports-props

ipip2005:
- in case the role is defined by expression or variable, means the role
   cannot be retrieved by tslint, it should pass the test


HamletDRC:
@t-ligu FYI, I will merge this. Let me know if you have any concerns.

PR #284, opened 2016-09-28
fix anchor fail when text length is exactly 4

ipip2005:
as well as test cases


src/reactA11yAnchorsRule.ts
```diff
@@ -58,7 +58,8 @@ class ReactA11yAnchorsRuleWalker extends ErrorTolerantWalker {
         while (this.anchorInfoList.length > 0) {
             const current: AnchorInfo = this.anchorInfoList.shift();
             this.anchorInfoList.forEach((anchorInfo: AnchorInfo): void => {
-                if (current.href === anchorInfo.href &&
+                if (current.href &&
```
ipip2005: Just want to mention that, IMO, if `href` is not defined, the `same href has same text` should not be applied

PR #297, opened 2016-10-20
skip a11y-role-supports-aria-props rule for custom element

### LINKED ISSUE #296: update rule: react-a11y-role-supports-aria-props - Do not check custom element
For example I have a custom element excepts React.HTMLProps
`<TextField aria-required={true} />`
The rule will fail because custom element has no role. Should skip custom rule for all these cases. Including other rules

ipip2005:
- also simplify some test-data file names
 - for react custom element e.g. <TextField />, skip checking the rule
   a11y-role-supports-aria-props
 - closes #296


ipip2005:
closes #296

PR #299, opened 2016-10-21
fix to make img element allow role='presentation' with non-empty alt

### LINKED ISSUE #298: update rule: react-a11y-img-has-alt - should allow role='presentation' with non-empty alt text.
There is a specific scenario that I pasted in the issue  https://github.com/Microsoft/tslint-microsoft-contrib/issues/298
we should allow role='presentation' appears with non-empty alt text. Like

```
<img
  aria-hidden={ true }
  role='presentation'
  src='...'
  alt='<non-empty> alt text'
/>
```
In following situation:
```

<img alt='123'/>
<span>123</span>.
```
Sometimes the alt text equals to the nearest text. The alt text is no need to be discoverable by assistive technologies. But they should exist to show for sighted user once this img fails loading.
For screen reader users, they only be able to 'see' one text around (e.g. there is a img with alt='apple' and there is a span text near it <span>apple</apple>, than screen reader users only see one of them)
For sighted users, they both see the img and the span, but in the situation that the img fails loading, sighted users should know `which image failed loading` by seeing the alt text at the position of that image.
This is for a better support for both screen reader users and sighted users.
So the following format should be acceptable:

```
<img aria-hidden='true' role='presentation' alt='123' />
<span>123</span>
```

ipip2005:
- consumers can now specify this options to declare whether there project supports Non-empty alt text with role equals presentation.
 - add one test case for the enabled option.
 - closes #298

### DIFF
--- a/src/reactA11yImgHasAltRule.ts
+++ b/src/reactA11yImgHasAltRule.ts
@@ -15,7 +15,7 @@ const ALT_STRING: string = 'alt';
 
 export function getFailureStringNoAlt(tagName: string): string {
     return `<${tagName}> elements must have an non-empty alt attribute or \
-use empty alt attribute and role='presentation' for presentational images. \
+use empty alt attribute as well as role='presentation' for decorative/presentational images. \
 A reference for the presentation role can be found at https://www.w3.org/TR/wai-aria/roles#presentation.`;
 }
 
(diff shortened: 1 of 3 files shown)


HamletDRC:
@bvancea @ipip2005 Why would you want to have an img element where role='presentation' and the alt is not empty? I thought that if an image was presentational then it was not visual and did not need to be read to the screen reader. Or is this required to support some combination of screenreader and browse?

ipip2005:
There is a specific scenario that I pasted in the issue  https://github.com/Microsoft/tslint-microsoft-contrib/issues/298
we should allow role='presentation' appears with non-empty alt text. Like

```
<img
  aria-hidden={ true }
  role='presentation'
  src='...'
  alt='<non-empty> alt text'
/>
```
In following situation:
```

<img alt='123'/>
<span>123</span>.
```
Sometimes the alt text equals to the nearest text. The alt text is no need to be discoverable by assistive technologies. But they should exist to show for sighted user once this img fails loading. So the following format is acceptable:

```
<img aria-hidden='true' role='presentation' alt='123' />
<span>123</span>
```

HamletDRC:
@ipip2005 My feeling on this is that your code will function correctly, but it is not a typical recommended best practice of a11y. On our project, for example, we would want all presentation images to *not* have an alt. 

Can we add a rule option called "allow-alt-with-role-presentation" and default the value to false? And then you can set it to true on your project?

ipip2005:
That is a good idea. I will following you suggestion.
Another off topic is that when we use such optional config for rules, we need more documentation against why we have this option. E.g in this case, we want to tell consumers in which case you should enable that, in which case you should not. If we don't, only few project consumers would like to enable/disable that specifically. I'm feeling the origin documentation in the home page is not meeting such requirements of long documentation.
If we have a plan of updating the documentation structure of this repo, I would love to help implement all a11y docs. Or if I had time, helping anything other I can do

PR #300, opened 2016-10-22
Remove deprecated no-duplicate-key rule

connor4312:
This has been deprecated in tslint's upcoming 4.x (https://github.com/palantir/tslint/issues/885) as it's functionality which is covered by the compiler. For that reason, removing the rule prior to release should have no negative effects and is one less thing to change down the line 😄

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -324,7 +324,8 @@ module.exports = function(grunt) {
             'no-unexternalized-strings': true,
             'object-literal-key-quotes': true,
             'no-relative-imports': true,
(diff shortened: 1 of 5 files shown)


HamletDRC:
@connor4312 do you know what version of TypeScript made this change? We support 1.5 forward. I am trying to figure out if 1.5, 1.6, or 1.7 needs this rule.

HamletDRC:
@connor4312 I just tested this and 1.5.3 will emit a compiler error for duplicate keys. I'll merge this in. 

Thanks for taking the time to send a PR, I really appreciate the effort.

PR #302, opened 2016-10-23
Port TSX curly spacing rule.

lijunle:
- The ESLint version: https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-curly-spacing.js
- The rule options are the same: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-curly-spacing.md

### DIFF
--- a/src/tsxCurlySpacingRule.ts
+++ b/src/tsxCurlySpacingRule.ts
@@ -1,0 +1,88 @@
+import * as ts from 'typescript';
+import * as Lint from 'tslint/lib/lint';
+
+/*
+ * TSX curly spacing rule.
+ *
+ * This is inspired by https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-curly-spacing.js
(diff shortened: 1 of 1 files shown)


HamletDRC:
hi @lijunle
Thanks for the rule and taking the time to submit this.

One note: eslint-plugin-react is an MIT licensed project. So if we copy and paste a regular expression from that project to ours then we need to declare the project in our copyright notice and give proper credit. I would prefer to have zero dependencies, so I will rewrite this regular expression. 

Another note: the build is failing because the rule is not documented in readme.md and there is no metadata. 

Lastly, I will add some unit tests. 

Thanks for the submission! I will try to merge this today.

HamletDRC:
Alright, I added tests and actually fixed an edge case. it all looks good. Thanks!

lijunle:
Super thank you for handling this! You are awesome!

HamletDRC:
Well, thanks for the PR! 

In the future, if you create a rule be sure to use: 

    grunt create-rule my-rule-name

Then it generates tests and documentation for you :)

PR #308, opened 2016-10-31
function-name rule - add support for linting protected methods

netroy:
this adds support for `protected` class methods on classes for the `function-name` rule.

also, removed the duplicate regexp across public, private & protected methods.


HamletDRC:
Thanks! I updated the readme as well.

PR #311, opened 2016-11-04
Configurable Props and State interface names

niik:
We currently use the tslint [`interface-name`](https://palantir.github.io/tslint/rules/interface-name/) rule to enforce a naming convention of all interfaces beginning with the capital letter `I` and we would very much like to use the `react-unused-props-and-state`.

Unfortunately, since that only looks at interfaces named `Props` or `State` it fails to pick up on our unused fields.

This PR adds an (optional) configuration for the rule which lets consumers pass a regular expression to match the names of their `Props` and `State` interfaces. We now use this to pass the following configuration:

```json
"react-unused-props-and-state": [ true,
  {
    "props-interface-regex": "Props$",
    "state-interface-regex": "State$"
  }
],
```

Which tells the rule to include all interfaces _ending_ in `Props` or `State`.


HamletDRC:
Thanks for the PR. I will definitely merge this, but I won't have time until tomorrow.

niik:
@HamletDRC No sweat, please let me know if there's anything I can do to help you get this merged

HamletDRC:
This is available on our "releases" branch now. You can point your npm dependency there if you need the latest files.

PR #318, opened 2016-12-05
Fix line endings errors in IDE by providing an .editorconfig file

bondz:
The editor should help out with line endings. Getting lots of expected lineending to be LF when testing.

VSCode, when using the Editor Config extension respects the `.editorconfig` file and saves files using the line ending specified. It can also recommend installing the extension if it has not been installed.


HamletDRC:
This is great! I also had these issues but didn't know to fix them this way. 

I am merging to the 3_0_0 branch

HamletDRC:
I marked it closed because we know we'll merge 3_0_0 to master.

PR #324, opened 2016-12-12
Reenable linting

bondz:
Linting of ts files was previously disabled. Grunt-tslint 4 now published supports tslint >= 4. Source linting has now been re-enabled.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -463,7 +463,7 @@ module.exports = function(grunt) {
         'copy:json',
         'ts',
         'mochaTest',
-        //'tslint', // commented out until grunt-tslint supports tslint 4.0
+        'tslint',
         'validate-documentation',
         'validate-config',
         'validate-debug-mode',
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -335,7 +335,12 @@ module.exports = function(grunt) {
             'no-inferrable-types': true,
             'only-arrow-functions': true,
             'ordered-imports': true,
-            'typedef-whitespace': true
+            'typedef-whitespace': true,
+            'array-type': true,
+            'completed-docs': true,
+            'cyclomatic-complexity': true,
+            'file-header': true,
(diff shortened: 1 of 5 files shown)


HamletDRC:
I don't understand why we got those false positives in those 2 rules. 
In my opinion, the semi-colon there is required and should not be removed. 

Do you have any ideas?

HamletDRC:
Also, I tried to make a unit test for these false positives and could not get the code to fail in the tests.

bondz:
The false positive, using the code

```ts
const FAILURE_STRING_FACTORY: (identifier: string) => string =
        (identifier: string) => `Identifier '${identifier}' never appears ` +
            'on the LHS of an assignment - use const instead of let for its declaration.'
```

vscode-tslint errors as appropriate for a missing semicolon.

But in the case of 
```ts
class Test {
    public static FAILURE_STRING_FACTORY: (identifier: string) => string =
        (identifier: string) => `Identifier '${identifier}' never appears ` +
            'on the LHS of an assignment - use const instead of let for its declaration.';

    public apply() {
        return 'test';
    }
}
```
vscode-tslint reports a false positive of an unnecessary semi-colon. This is a bug with the rule

HamletDRC:
The weird part is that I can put this code into a unit test and the test does not fail :(

bondz:
That's weird.

HamletDRC:
@bondz I solved the mystery. The false positive is a bug in the tslint "semicolons" rule, not a bug in the tslint-microsoft-contrib no-unnecessary-semicolons rule. I created a tslint issue: https://github.com/palantir/tslint/issues/1895

bondz:
Glad you tracked it down @HamletDRC 😌 That's a relief.

bondz:
@HamletDRC with recent changes, semi-colons are no longer required 😞 and linting for extraneous semi-colons throw a `missing-whitespace` error instead.

bondz:
Yes.

PR #345, opened 2017-02-16
Remove no-unused-variable to suppress tslint warning

yokomotod:
Now tslint shows warning if we load `no-unused-variable` even it is false.

In tslint side, it is already removed from recommended config. ( palantir/tslint#1748 )


LukeSkyw:
Problem still here with:
```
"tslint": "^4.4.2",
"tslint-microsoft-contrib": "^4.0.0",
```

HamletDRC:
I just started the 4.0.1 release process. It will hit npm within 24 hours. Thanks for waiting!

PR #347, opened 2017-03-16
Ensure LF line endings are used upon git checkout.

chrismbarr:
Getting latest from master and simply running `grunt all` on my Windows 10 machine produced **a lot** of linting errors about line endings

lots of errors

 quick google brought me to [this page](https://help.github.com/articles/dealing-with-line-endings), where I added the line line to the `.gitattrubutes` file (which is the only change in this PR) & then ran the commands to refresh my local repo with the new settings, and now I get zero errors.

no errors no

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 * text=auto
+* text eol=lf


bgold09:
@chrismbarr wanted to let you know that the appveyor build is failing because it is brand new. My PR #350 is addressing this by introducing a proper appveyor configuration; until then, you can safely ignore these failures.

HamletDRC:
This change doesn't actually fix the problem, right? When you initially clone the repo there is not yet a .gitattributes file so you'll get the files using your default line endings. This is why the README asks that you clone the repo with this command: 

    git clone git@github.com:Microsoft/tslint-microsoft-contrib.git --config core.autocrlf=input --config core.eol=lf

I guess I can commit this, but it doesn't really fix the problem when cloning the repo.

HamletDRC:
In other words, I think your changes here will "Ensure LF line endings are used upon a git pull" but they will not "Ensure LF line endings are used upon git checkout".

chrismbarr:
Ah ok, that makes sense.  My process here was to:

1. pull
2. get lint errors
3. make the commit in this PR
4. follow [these instructions](https://help.github.com/articles/dealing-with-line-endings/)
    - run `git rm --cached -r .`
    - run `git reset --hard`

Doing all of that worked for me.

HamletDRC:
If you don't clone with the right parameters then you're going to need to do the --cached --hard thing regardless.

PR #348, opened 2017-03-16
Fix the new rule snippet to not create an immediately broken file

chrismbarr:
I wanted to create a new rule for this project, so I followed the instructions and ran
```
grunt create-rule --rule-name=my-new-rule
```
which created a broken rule file that I had to figure out why it was broken before I could start developing.  This PR should fix the incorrect path to `tslint` and add the missing required properties to the `metadata` class member.

Perhaps some tests should be added for this project that creates a rule with this command and ensures that it can be compiled?


HamletDRC:
Thanks for noticing and fixing!

PR #361, opened 2017-04-09
Use TypeScript config files for compilation (fixes #360)

### LINKED ISSUE #360: gruntfile and tsconfig are inconsistent
The grunt task for compiling the TypeScript source and the tsconfig.json in this repo do not match; clearly the config defined for grunt is the one being used in CI builds. I'd propose we keep configurations inside tsconfig.json where most people expect to find the compiler options, and let the grunt task use this file. That way, compilation is consistent regardless of whether using grunt of tsc directly. The tsconfig.json also is how VS Code finds the options to use and I'm not sure it can integrate with grunt in that way.

@HamletDRC what would you prefer?

bgold09:
* Adding two TypeScript configuration files for compiling TypeScript files: tsconfig.json for files under "src" and tsconfig.testdata.config for files under "test-data".
  * Grunt will use these files in its "ts" tasks. Grunt file globbing is used to discover the files to compile, and will write the lists of files into the tsconfig files. This is used instead of the include/exclude support for tsconfigs because that feature does not behave as expected and will still include some other files that    should not be compiled.
  * This also fixes a consistency issue when using VS Code, since the IDE will try to use a tsconfig in the root of the workspace; previously this would bring up inconsistent results since the true compilation options were in the grunt configuration.


HamletDRC:
@bvancea does this tsconfig configuration with all the listed files look normal to you?

bgold09:
@HamletDRC, @bvancea: are there any concerns with this? My inbound PR for making these rules compatible with TSLint 5.0.0 are partially dependent on this change

loicraux:
@bgold09 Why do we now list all the files in the `tsconfig.json` file ?
I read your comment but it is not clear to me why this is needed instead of a glob pattern ?
Do this mean that whenever a new rule is added, the `tsconfig.json` file must be edited ?
Thanks !

PR #375, opened 2017-05-23
Only return expression length to stop annoying whole block underlining

startswithaj:
While developing you often want to select one set of tests to run. In vs code because mocha-avoid-only returns the entire length of the block, every single line gets underlined, making it difficult to ignore and dev against.

screen shot 2017-05-23 at 10 57 12 am

vs

screen shot 2017-05-23 at 10 58 10 am


HamletDRC:
This is small but super valuable. Thanks for taking the time to clean this up!

PR #377, opened 2017-06-01
Add a fixer for the no-var-self rule

sethbrenith:
This change adds automatic fixing for `no-var-self` if a few criteria are met:

1. The variable declaration takes place within a function, method, or constructor
2. The variable declaration does not specify a type
3. The variable is declared on its own, not in a list of variables
4. No other declaration of another variable with the same name takes place within the same function or any nested scope within it
5. There are no assignments to the variable after the declaration
6. There are no uses of the variable within nested functions defined with the `function` keyword (arrow syntax is allowed because it preserves the outer `this`)

### DIFF
--- a/src/noVarSelfRule.ts
+++ b/src/noVarSelfRule.ts
@@ -1,6 +1,7 @@
 import * as ts from 'typescript';
 import * as Lint from 'tslint';
 
+import { getNextToken, getPreviousToken } from 'tsutils';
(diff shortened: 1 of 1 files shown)


sethbrenith:
Oops, turns out I can't just plop code that I developed in `tslint` into `tslint-microsoft-contrib` and assume it will compile. Sorry!

PR #386, opened 2017-08-04
Add 'typescript' to peer dependencies

rogierschouten:
Most files import "typescript" but typescript isn't depended on in package.json.  You usually don't notice because typescript is installed by other modules (e.g. tslint), but it isn't proper.

Like tslint, I put typescript in the peerDependencies and I used the version range that tslint 5.1 also has in its peerDepdenencies for typescript.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "underscore": "1.8.3"
   },
   "peerDependencies": {
-    "tslint": ">=5.1.0",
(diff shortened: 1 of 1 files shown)


HamletDRC:
@rogierschouten we don't actually test the project on all those typescript versions and point releases are often not backwards compatible at the AST level. But it is no different than not declaring a dependency, I suppose. 

Can we just declare anything below 3.0 as the dependency? It is simpler in the dependency declaration and is more future safe. Do you see an issue with that?

rogierschouten:
@HamletDRC that's fine I think, although this may lose some -dev versions I guess. I've changed it both for TypeScript and TSLint.

PR #391, opened 2017-09-03
#391 Upgrade grunt and other library versions in build

dawsbot:
* Previous version of grunt fails on raw `npm install && npm test` if you don't have grunt installed globally.

This makes it possible for a fresh contributor (like me!) to get downloaded and passing tests instantly

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,9 +8,18 @@
   },
   "author": "Hamlet D'Arcy",
   "contributors": [
-    { "name": "Bernd Kiefer",   "email": "bernd.kiefer@microsoft.com" },
-    { "name": "Daniel Manesku", "email": "daniel.manesku@microsoft.com" },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -8,9 +8,18 @@
   },
   "author": "Hamlet D'Arcy",
   "contributors": [
-    { "name": "Bernd Kiefer",   "email": "bernd.kiefer@microsoft.com" },
-    { "name": "Daniel Manesku", "email": "daniel.manesku@microsoft.com" },
-    { "name": "Hamlet D'Arcy",  "email": "hamlet.darcy@microsoft.com" }
+    {
```
dawsbot: Auto-formatted by npm 5 (I did not include the lockfile in this PR since y'all don't use npm 5 on this project)

HamletDRC:
thanks, I appreciate you taking the time to do this.

PR #399, opened 2017-10-19
Use ReadonlyArray for node arrays

bowenni:
TypeScript 2.5 (Microsoft/TypeScript#17213) specified the NodeArray type to be readonly. This doesn't change the run-time behavior but will cause compile-time errors if compiling against ts2.5.
Changing types for node arrays to ReadonlyArray fixes the problem.


HamletDRC:
@bowenni  can you please click the "sign now" button? It is a Contributor License Agreement.

PR #402, opened 2017-11-16
fix FP for no-string-based-set-timeout

tsmolka:
I found this false positive when running `no-string-based-set-timeout` over my code base. Proposed patch fixed it in my case.


HamletDRC:
Nice, thanks!

lijunle:
Hi, @HamletDRC What is the plan to release version 5.0.2? We are enabling `no-string-based-set-timeout` but see this false-positive.

HamletDRC:
doing it now :)

lijunle:
Just saw it! Thank you!

default

tsmolka:
Thank you!

PR #406, opened 2017-12-14
fix typo

hrldcpr:
"If you ~~reply~~rely on version ranges in your dependencies"


HamletDRC:
@hrldcpr thank you for taking the time to fix this for us.

PR #408, opened 2017-12-18
#408 un-deprecate some rules

HamletDRC:
function-name / variable-name rule.
 - we allow much more configuration than tslint.
 - for example, you can define a convention for privates, and a different convention for statics

no-reserved-keywords / variable-name
 - the variable-name rule is not enough to pass the tests for no-reserved-keywords
 - we have more keywords and more usages

missing-jsdoc / file-header
 - missing-jsdoc makes sure you have a top level JSDoc comment somewhere. comment could be at start of file, start of module, or start of class
 - file-header makes sure you have a top level text at start of file

no-unnecessary-field-initialization / TSLint no-unnecessary-initializer
 - these are actually different rules
 - tslint analyzes everything *except* class members: https://github.com/palantir/tslint/blob/master/src/rules/noUnnecessaryInitializerRule.ts
 - our rule analyzes *only* class members

prefer-array-literal / TSLint array-type
 - these are not drop in replacements. Our rule does a fair bit more than the base TSLint rule.

prefer-type-cast / no-angle-bracket-type-assertion
 - These rules are actually opposites of each other
 - TSLint claims to be opinionated, so if they don't like a rule then they don't accept it into their ruleset
 - They told me they don't want this rule in the base product

promise-must-complete / TSLint no-floating-promises
 - These rules are coded fairly differently, and are not replacements
 - It appears that the tslint rule does not satisfy the unit tests for promise-must-complete.
 - no-floating-promises requires type checking

react-tsx-curly-spacing / tslint-react jsx-curly-spacing rule.
 - I generally don't recommend tslint-react over tslint-microsoft-contrib
 - There is a ton of overlap between the two projects


HamletDRC:
@JoshuaKGoldberg Please take a look at this pull request. I think some of these rules should not have been deprecated.

JoshuaKGoldberg:
Hey sorry this slipped off my radar. LGTM

PR #411, opened 2018-01-22
Removed default configuration of deprecated rules.

### LINKED ISSUE #410: 5.0.2 gives warning about deprecated rules
Using tslint with 5.0.2 I get this:

Warning: no-duplicate-case rule is deprecated. Replace your usage with the TSLint no-duplicate-switch-case rule.
Warning: no-var-self rule is deprecated. Replace your usage with the TSLint no-this-assignment rule.
Warning: valid-typeof rule is deprecated. Replace your usage with the TSLint typeof-compare rule.

Seems these rules were removed in the main part of the recommended rules set, but then they are turned on again at the end:

        /**
         * Deprecated rules.  The following rules are deprecated for various reasons.
         */
        "missing-optional-annotation": false,  // now supported by TypeScript compiler
        "no-duplicate-case": true,
        "no-duplicate-parameter-names": false, // now supported by TypeScript compiler
        "no-empty-interfaces": false, // use tslint no-empty-interface rule instead
        "no-missing-visibility-modifiers": false, // use tslint member-access rule instead
        "no-multiple-var-decl": false,         // use tslint one-variable-per-declaration rule instead
        "no-stateless-class": true,
        "no-switch-case-fall-through": false,  // now supported by TypeScript compiler
        "no-var-self": true,
        "react-tsx-curly-spacing": true,
        "typeof-compare": false,               // the valid-typeof rule is currently superior to this version
        "valid-typeof": true,

caghand:
Fixes #410.
I could have set all deprecated rules to "false".  But instead, I have dropped them completely.  I did not want to interfere with any rule configurations that the user might have customized elsewhere.


caghand:
I had to go back to my original idea of setting all deprecated rules to "false".  Otherwise, validate-config is failing because rules are missing from config.  In the next major revision, ideally, all of these deprecated rules should be scrubbed from the rule repository completely.

HamletDRC:
@JoshuaKGoldberg  @caghand 
Do you need a release? 

Also, check out the final commit. The recommended ruleset is generated by the source file, so you turn rules off in the source code of that rule.

HamletDRC:
final commit is here: https://github.com/Microsoft/tslint-microsoft-contrib/commit/ed832e755845240d71a400b6a75e5251be7777c4

caghand:
Thank you for the fix! :)
Well, it would be nice to have a release, because we would be able to get rid of the warnings in our build.  It would fix #410 basically.
Are you planning to release 5.0.3 soon?

vreality64:
Is there any news about 5.0.3 release ? I'm waiting for version 5.0.3 ~ 😃

appsolutegeek:
me too :-) I notice I have the following being displayed. I presume that this commit fixes that - right ?

Warning: no-duplicate-case rule is deprecated. Replace your usage with the TSLint no-duplicate-switch-case rule.
Warning: no-stateless-class rule is deprecated. Replace your usage with the TSLint no-unnecessary-class rule.
Warning: no-var-self rule is deprecated. Replace your usage with the TSLint no-this-assignment rule.
Warning: valid-typeof rule is deprecated. Replace your usage with the TSLint typeof-compare rule.

caghand:
@vreality64 and @appsolutegeek: Version 5.0.3 has been released.  And you are right, this commit (and the associated release) fixes the warnings you are seeing.

PR #417, opened 2018-02-13
proposed type-checking for noStringBased rules

tsmolka:
I experienced false positives in `no-string-based-set-timeout` rule when `setTimeout` is called on [http.ClientRequest](https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback) from module `http`. 

In this PR I am proposing to add additional type checks so that such FPs are eliminated. Existing tests are successful but we might need to test this on larger code base.


HamletDRC:
I tested this on Microsoft Social Engagement codebase and it fixed a couple false positives we had. No errors that I saw.

PR #420, opened 2018-03-21
Fixed no-unnecessary-local-variable for recursive variable references

### LINKED ISSUE #381: no-unnecessary-local-variable false positive
Code like the following incorrectly triggers a lint error:

```
const balls = {
    foo: 'foo',
    bar: () => balls.foo
};

return balls;
```

JoshuaKGoldberg:
If a variable references itself in its own initialization, the rule will no longer complain about it.

Fixes #381


JoshuaKGoldberg:
I also fixed up casts, but didn't change from `null` to `undefined`.

PR #421, opened 2018-04-17
Dedupe SUSPICIOUS_WORDS list for the no-suspicious-comment rule

astorije:
This reduces potential out-of-sync issue if one gets updated without the other, or if one day the list can be configured.


astorije:
Argh yes, I realized this when I was not on the laptop anymore... Sorry!! 🤦‍♂️

PR #423, opened 2018-04-19
Allow HTTP scheme for localhost URLs

astorije:
In my experience, localhost URLs are almost exclusively used for development purposes, and setting up HTTPS for localhost URLs is really tedious, and most of the time unnecessary.

Note that I'm not completely sure of the syntax, so let me know if I'm wrong! The `:` at the end is on purpose to make sure all ports are allowed. I don't think it's worth the effort to allow `http://localhost/` without a port specifically, but let me know.

screen shot 2018-04-19 at 17 59 29

(On an unrelated note, I suggest removing `http://examples.com`, it seems like a semi-real domain, and it's not part of https://en.wikipedia.org/wiki/Example.com / https://tools.ietf.org/html/rfc2606#section-3)

### DIFF
--- a/recommended_ruleset.js
+++ b/recommended_ruleset.js
@@ -19,7 +19,7 @@ module.exports = {
         "no-eval": true,
         "no-exec-script": true,
         "no-function-constructor-with-string-args": true,
-        "no-http-string": [true, "http://www.example.com/?.*", "http://www.examples.com/?.*"],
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
SGTM, thanks for the add!

I personally work in a setup that does care about the difference, but I agree that in the majority of cases, this is fine.

If you could send another PR for the examples removal that'd be great! I can if you don't have the time.

astorije:
> If you could send another PR for the examples removal that'd be great! I can if you don't have the time.

No worries, done in #449, thanks for reaching out! ❤️

PR #427, opened 2018-05-02
Disable the use of `console.error`

astorije:
I have noticed that this project already locally disables this rule (as if it were present in the config) [here](https://github.com/astorije/tslint-microsoft-contrib/blame/master/src/functionNameRule.ts#L101-L103) and [here](https://github.com/astorije/tslint-microsoft-contrib/blame/master/src/reactUnusedPropsAndStateRule.ts#L71-L73) (the only 2 uses of `console.error` in this repo).

We realized `console.error` was not forbidden in the recommended configuration and think it would be a sane thing to enforce by default. I reckon this will be a breaking change however.

### DIFF
--- a/recommended_ruleset.js
+++ b/recommended_ruleset.js
@@ -48,7 +48,7 @@ module.exports = {
         "no-backbone-get-set-outside-model": true,
         "no-bitwise": true,
         "no-conditional-assignment": true,
-        "no-console": [true, "debug", "info", "log", "time", "timeEnd", "trace"],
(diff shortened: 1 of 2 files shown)


astorije:
Hey there, any chance someone could take a look at this PR (and other open PRs)? 🙂
Thanks in advance!!

JoshuaKGoldberg:
Hey @astorije, sorry about the delay! Agreed that this should be the case (and that it's a breaking change). Should be noted in the next release.

PR #428, opened 2018-05-02
Only flag html method if jQuery (#425)

### LINKED ISSUE #425: no-inner-html rule: false positive for non-jQuery scenario
I have a project using:
* enzyme
* chai
* chai-enzyme
* mocha-webpack

I have something similar to the following in a test:
```javascript
// $ cat test/AuthenticatedApp.test.tsx
import React from 'react';                                                                                                                                               
import { expect } from 'chai';                                                                                                                                           
import { render } from 'enzyme';
import AuthenticatedApp from '../src/AuthenticatedApp.tsx';

describe('AuthenticatedApp', function() {
  context('first', function() {
    it('renders children', function() {
      const app = render(
        <AuthenticatedApp session={{}}>
          <h1>Hi guys</h1>
        </AuthenticatedApp>,
      );
      expect(app).to.have.html('<h1>Hi guys</h1>');  // ERROR
    });
  });
});
```

I get the following error:
```
test/AuthenticatedApp.test.tsx14, 7]: Using the html() function to write a string to innerHTML is insecure: expect(app).to.have.html('<h1>Hi guys</h1>')
```

This is incorrect behavior. I am not using the [.html](https://api.jquery.com/html/) function, I am using a [chai-enzyme assertion](https://github.com/producthunt/chai-enzyme#htmlstr)

I believe this could be fixed by adding an `isJquery` check alongside https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/noInnerHtmlRule.ts#L57

Workaround: I have disabled `no-inner-html` in my `test` directory.

amacleay:
* Add unit tests for isJQuery method
* Add new method, `isJQueryExpression`, which returns true if the
  expression is a result of a jQuery function call or a function
  expression on a variable whose name starts with `$`, which is a common
  convention for identifying variables which are the result of a jQuery
  expression; add tests for it
* Modify `no-inner-html` to only flag method calls if the method
  invocant appears to be a jQuery expression

Resolves #425

### DIFF
--- a/src/noInnerHtmlRule.ts
+++ b/src/noInnerHtmlRule.ts
@@ -7,7 +7,7 @@ import {ExtendedMetadata} from './utils/ExtendedMetadata';
 
 const FAILURE_INNER: string = 'Writing a string to the innerHTML property is insecure: ';
 const FAILURE_OUTER: string = 'Writing a string to the outerHTML property is insecure: ';
-const FAILURE_JQUERY: string = 'Using the html() function to write a string to innerHTML is insecure: ';
+const FAILURE_HTML_LIB: string = 'Using the html() function to write a string to innerHTML is insecure: ';
 
 /**
  * Implementation of the no-inner-html rule.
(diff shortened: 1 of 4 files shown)


astorije:
@HamletDRC, ping? :)

JoshuaKGoldberg:
Thanks for the PR @amacleay! I posted some feedback on the linked issue.

amacleay:
New commit makes the matcher configurable via an option named `jquery-matcher`

amacleay:
`html-lib-matcher` and `htmlLibExpressionRegex` seem alright? `html-matcher` to me suggests it's a regex for matching an expression which is HTML.

astorije:
> For example, if a project only uses Cheerio, they might not know or feel comfortable using a config option that only references jQuery.

I personally don't have a preference on naming, but note that even [Cheerio](https://github.com/cheeriojs/cheerio) mentions it's an implementation of core jQuery, so that would probably not sound unfamiliar to Cheerio users.

astorije:
@JoshuaKGoldberg, do you need anything else to be able to merge this?

Also, any chance you could make a release? We'd love to consume a published version of this :)

JoshuaKGoldberg:
Ah my apologies, the only blocker was me forgetting to hit a button.

For a release, sure - I've only recently become a maintainer of the repository so I'll schedule some time with @HamletDRC soon™ to get that going!

Issue #339, opened 2017-01-29
react-unused-props-and-state fails with SFC's


bookman25:
This rule appears to only work with `this.props` calls.  So SFC's don't work because there is no `this` pointer.

Also, once you destructure props/state, it assumes all keys are used.  Even if you only reference some of the props in the destructure:

```
interface State {
  key1: string;
  key2: string;
}
...
const { key1 } = this.state;  // this will mark all keys as used
```

HamletDRC:
Can you post a sample of one of your SFCs?

HamletDRC:
Here is one possible syntax for SFC:

```
import classNames = require('classnames');
import CommonKey = require('common/locales/CommonKey');
import I18nFacade = require('common/i18n/I18nFacade');
import React = require('react');

'use strict';

const buttonLabel: string = I18nFacade.TEXT(CommonKey.close());

module XCloseButton {

    export interface Props extends React.Props<React.HTMLAttributes> {
        className?: string;
        onClick: () => void;
    }
}

/**
 * Render an 'X' styled closed button
 */
/* tslint:disable:variable‐name */
function XCloseButton({ className, onClick }: XCloseButton.Props): ReactTypes.ReactElement<any> {
/* tslint:enable:variable‐name */
    return (
        <button
            className={classNames('XCloseButton', className)}
            onClick={onClick}
            aria-label={buttonLabel}
            title={buttonLabel}
        />
    );
}

export = XCloseButton;
```

narthollis:
I am also experiencing this issue, `tslint` output follows reproduction.

#### Reproduction
```typescript
interface Props {
    children?: React.ReactNode;
    heading: string;

    className?: string;
}

export function TestComponent(props: Props) {
    return (
        <div className={props.className || ''}>
            <h3>{props.heading}</h3>
            {props.children}
        </div>
    )
}

export default TestComponent;
```

#### Output
```
src/components/TestComponent.tsx
58:6  semicolon                     Missing semicolon
46:5  react-unused-props-and-state  Unused React property defined in interface: children
47:5  react-unused-props-and-state  Unused React property defined in interface: heading
49:5  react-unused-props-and-state  Unused React property defined in interface: className
```

bookman25:
Sorry about the late response, just saw there was a comment.  Here's an example SFC.  Both properties are being flagged as unused.

__tslint.json__
```
"react-unused-props-and-state": [
	true,
	{
		"props-interface-regex": "IProps$",
[...]

PR #431, opened 2018-06-04
fixing unused props and state for SFC components

mikea:
Very crude fix for #339 
I'm sure it doesn't handle all cases, but this is something to start with.


JoshuaKGoldberg:
Hey I'm going to close this PR for housekeeping. Feel free to continue discussing this in the issue!

PR #443, opened 2018-06-20
export-name: detect export declarations

### LINKED ISSUE #444: export-name: export declarations are not flagged
Suppose I have a file named `passLint.ts`:
```typescript
// $ cat src/failLint.ts
export function fish() : string {
  return '🐟';
}
```

The `export-name` rule will rightly complain that the export name does not match.

Suppose that, instead, I write the export on a different line from the function declaration:
```typescript
// $ cat src/passLint.ts
function fish() : string {
  return '🐟';
}
export { fish }
```

`export-name` fails to fire on this, even though it results in identical javascript output:
```javascript
// $ tsc; cat src/failLint.js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
function fish() {
    return '🐟';
}
exports.fish = fish;

// $ diff src/*.js  # produces no output because files match
```

---
All source code for a minimal example:
```
$ for x in $( git ls-files | grep -v yarn ); do echo;echo "# $x"; cat $x; done

# .gitignore
node_modules
yarn-error.log

# package.json
{
  "name": "tmp.xj0vgsu2jj",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "lint": "tslint --format stylish --project ."
  },
  "author": "",
  "license": "UNLICENCED",
  "devDependencies": {
    "tslint": "^5.10.0",
    "tslint-microsoft-contrib": "^5.0.3",
    "typescript": "2.9.2"
  }
}

# src/failLint.ts
export function fish() : string {
  return '🐟';
}

# src/passLint.ts
function fish() : string {
  return '🐟';
}
export { fish };

# tsconfig.json
{
  "compilerOptions": {
    /* Basic Options */
    "target": "ES5",
    "lib": [ // Specify library files to be included in the compilation.
      "es2015",
      "es2017"
    ],

    /* Strict Type-Checking Options */
    "strict": true, // Enable all strict type-checking options.
    "strictNullChecks": true, // Enable strict null checks.

    /* Additional Checks */
    "noUnusedLocals": true, // Report errors on unused locals.
    "noUnusedParameters": true, // Report errors on unused parameters.
    "noImplicitReturns": true, // Report error when not all code paths in function return a value.
    "noFallthroughCasesInSwitch": true, // Report errors for fallthrough cases in switch statement.

    /* Module Resolution Options */
    "moduleResolution": "node", // Specify module resolution strategy.
    "allowSyntheticDefaultImports": true, // Allow default imports from modules with no default export. This does not affect code emit, just typechecking.
    "esModuleInterop": true, // Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports.
  },
  "exclude": ["node_modules/"]
}

# tslint.yml
extends:
  - tslint-microsoft-contrib # https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/recommended_ruleset.js
rules:
  missing-jsdoc: false
```

amacleay:
Previously, the `export-name` rule would detect an export assignment
like

```
export = Fish;
// or
export default Fish;
```

but it would not detect an export like

```
export { Fish }
```

Fixes #444

### DIFF
--- a/src/exportNameRule.ts
+++ b/src/exportNameRule.ts
@@ -46,26 +46,43 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 }
 
+function isExportedDeclaration(element: ts.Statement): boolean {
+    return AstUtils.hasModifier(element.modifiers, ts.SyntaxKind.ExportKeyword);
+}
+
+type ExportStatement = ts.ExportDeclaration | ts.ExportAssignment;
+function isExportStatement(node: ts.Statement): node is ExportStatement {
+  return ts.isExportAssignment(node) || ts.isExportDeclaration(node);
+}
+
+function getExportsFromStatement(node: ExportStatement): [string, ts.Node][] {
+  if (ts.isExportAssignment(node)) {
+      return [[node.expression.getText(), node.expression]];
+  } else {
+      const symbolAndNodes: [string, ts.Node][] = [];
+      node.exportClause.elements.forEach(e => {
+          symbolAndNodes.push([e.name.getText(), node]);
+      });
+      return symbolAndNodes;
+  }
+}
+
 export class ExportNameWalker extends ErrorTolerantWalker {
     protected visitSourceFile(node: ts.SourceFile): void {
 
         // look for single export assignment from file first
-        const singleExport: ts.Statement[] = node.statements.filter((element: ts.Statement): boolean => {
-            return element.kind === ts.SyntaxKind.ExportAssignment;
-        });
+        const singleExport = node.statements.filter(isExportStatement);
         if (singleExport.length === 1) {
-            const exportAssignment: ts.ExportAssignment = <ts.ExportAssignment>singleExport[0];
-            this.validateExport(exportAssignment.expression.getText(), exportAssignment.expression);
+            const symbolsAndNodes = getExportsFromStatement(singleExport[0]);
+            if (symbolsAndNodes.length === 1) {
+                this.validateExport(symbolsAndNodes[0][0], symbolsAndNodes[0][1]);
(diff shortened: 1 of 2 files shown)


src/tests/ExportNameRuleTests.ts
```diff
@@ -315,6 +315,24 @@ describe('exportNameRule', () : void => {
             ]);
         });
 
+        it('when mis-named function is exported in a separate statement', () : void => {
```
JoshuaKGoldberg: Good test for the case of complaining. There should also be a corresponding test for the case of the user getting it right, e.g. `file.ts` having `export { file };`.

src/exportNameRule.ts
```diff
@@ -46,26 +46,43 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 }
 
+function isExportedDeclaration(element: ts.Statement): boolean {
```
JoshuaKGoldberg: Just a suggestion: you may wish to file issues on https://github.com/ajafff/tsutils to add these there, as they seem useful. Not a blocking issue for this pull request at all.

amacleay: You got it: https://github.com/ajafff/tsutils/issues/68 - feel free to comment there if it's not clear or not quite what you expected.

src/tests/ExportNameRuleTests.ts
amacleay: Added in new commit, good catch!

JoshuaKGoldberg:
Awesome.

Technically, this is a breaking change, as it should start flagging violations it wasn't before.

PR #447, opened 2018-07-03
Update README due to broken link

jpinho:
Updates broken link for `react-a11y-props`


JoshuaKGoldberg:
Nice find, thanks @jpinho!

PR #460, opened 2018-07-20
Issue-207 allow second parameter of expect to satisfy documentation rule

### LINKED ISSUE #207: chai-vague-errors should check 2nd expect arg for an error description
It's possible to [specify a custom error message](http://chaijs.com/guide/styles/#expect) via the second arg to `expect()`

``` js
// AssertionError: topic [answer]: expected 43 to equal 42.
expect(43, 'topic [answer]').to.equal(42);
```

But `chai-vague-errors` produces `Found chai call with vague failure message. Please add an explicit failure message`

With the 2nd arg messages, we can continue to use the fluent chai syntax:

``` js
expect(something, 'expected something to have occurred').to.be.true;
```

without having to resort to:

``` js
expect(something).to.equal(true, 'expected something to have occurred');
```

So, should `chai-vague-errors` accommodate 2nd arg messages?

pburrows:
(first PR... hopefully did it right). Hopefully this fixes this issue:

Fixes #207


src/tests/ChaiVagueErrorsRuleTests.ts
```diff
@@ -269,7 +280,7 @@ describe('chaiVagueErrorsRule', () : void => {
         ]);
     });
 
-    it('should fail on strictly equality in expectation', () : void => {
+    it('should fail on strict equality in expectation', () : void => {
```
JoshuaKGoldberg: Nice catch 😄

recommended_ruleset.js
```diff
@@ -133,7 +133,7 @@ module.exports = {
         "no-for-in": true,
         "no-function-expression": true,
         "no-inferrable-types": false, // turn no-inferrable-types off in order to make the code consistent in its use of type decorations
-        "no-multiline-string": false,
+        "no-multiline-string": true, // multiline-strings often introduce unnecessary whitespace into the string literals
```
JoshuaKGoldberg: Looks like some unrelated changes on this file made it in. Could you please undo/revert changes to `recommended_ruleset.js` and `package-lock.json`?

Separately, I think we should start checking in `package-lock.json`, so I've filed #462 to do that.

src/chaiVagueErrorsRule.ts
```diff
@@ -41,7 +41,12 @@ class ChaiVagueErrorsRuleWalker extends ErrorTolerantWalker {
     protected visitPropertyAccessExpression(node: ts.PropertyAccessExpression): void {
         if (ChaiUtils.isExpectInvocation(node)) {
             if (/ok|true|false|undefined|null/.test(node.name.getText())) {
-                this.addFailureAt(node.getStart(), node.getWidth(), FAILURE_STRING);
+                const expectInvocation: ts.CallExpression = ChaiUtils.getExpectInvocation(node);
+                if (expectInvocation && expectInvocation.arguments.length === 2) {
```
JoshuaKGoldberg: Style nitpick: if you used `!expectInvocation || expectInvocation.arguments.length !== 2`, you could get rid of the `// do nothing` and `} else {` lines and just have this be one `if` statement. Not a blocker at all.

src/tests/ChaiVagueErrorsRuleTests.ts
```diff
@@ -17,6 +17,17 @@ describe('chaiVagueErrorsRule', () : void => {
         TestHelper.assertViolations(ruleName, script, [ ]);
     });
 
+    it('should pass on xxx fluent', () : void => {
+        const script : string = `
+            expect(something, 'message').to.be.true;;
+            expect(something, 'message').to.be.false;;
+            expect(something, 'message').to.not.be.null;;
+            expect(something, 'message').to.not.be.undefined;;
```
JoshuaKGoldberg: Style nitpick: any reason for `;;` instead of `;` here? Not a blocker at all, just sticks out a teeny bit.

pburrows: I'll remove it, but the original passing tests (`should pass on xxx`) had them, so I copied that style. Not sure why they were there in the first place.

JoshuaKGoldberg:
Awesome, thanks!

Issue #481, opened 2018-08-06
Documentation for additional_rule_metadata.json / recommended_ruleset.js.


JoshuaKGoldberg:
`recommended_ruleset.js` is actually auto-generated from `additional_rule_metadata.json` and checked at build-time by the `validate-config` task. This process should be documented: preferably also in the files it touches, so it's clear where the sources of truth for configs are.

JoshuaKGoldberg:
☠️  *It's time!* ☠️ 

Per #876, this repository is no longer accepting feature pull requests. TSLint is being deprecated and we recommend you switch to https://typescript-eslint.io.

Thanks for open sourcing with us, everyone!

Issue #454, opened 2018-07-05
Consider taking no-multiline-string out of the recommended preset


astorije:
We use and love the recommended preset at @CloudHealth, but we find `no-multiline-string` to be an unnecessarily restrictive rule to recommend by default.

As far as I can tell, there is nothing inherently bad about multiline template strings, and they're not one of those things accepted to be dangerous or recommended not to use by the community such as `eval` or `==`.

In our case, and this is probably the biggest reason why we disabled it (but not the only one), GraphQL queries are most of the time represented using multiline template strings:

```graphql
query UserQuery($email: string{
  user(id: $email) {
    name
    avatar
  }
}
```

This is true with both Apollo Client and Relay Client.

We also find this useful when building CLIs for example, but it's less easily found across our codebase.

So there is it, I suggest removing it from the recommended preset, from the point of view of a consumer :)
Thanks and keep up the good work!

JoshuaKGoldberg:
This sounds reasonable (I too always disable it 😃). Let's wait a bit and see if there are any voices against it, and if in a few weeks time nobody has piped up, it sounds like a good `Good First Issue 🙌`.

astorije:
> Let's wait a bit and see if there are any voices against it

That's totally reasonable. I opened #456 in preparation for it, so that when/if no one objects, it can be merged, or close if someone objects.
Thanks for the feedback @JoshuaKGoldberg!

PR #482, opened 2018-08-06
Finished removing no-multiline-string from recommended ruleset

JoshuaKGoldberg:
Opens #481. Continues on #454.

Turns out there's a bit more work that should be documented.


astorije:
Wow, sorry I missed all that!!

Issue #486, opened 2018-08-09
newline-before-return in recommended preset


astorije:
Hi there!

Like I mentioned before, we use and love this package and recommended ruleset, and I know we'll contribute more to it in the coming months.

One rule we dislike is `newline-before-return`. We think it does not bring any value (as opposed to padding blocks overall, which ESLint allows but not TSLint), and it even ends up in silly situations when you try to leverage FP-style methods (`.map`, etc.), for example:

```ts
foo.reduce((acc, item) => {
  acc[item] = '💥';

  return acc;
}, {});
```

In fact, I noticed that this repo itself calls it ["kind of a silly rule"](https://github.com/Microsoft/tslint-microsoft-contrib/blob/4de93cf8ce385adf7ca4315d9841a0b801b175db/Gruntfile.js#L328) 😁

On top of this, this rule is not auto-fixable, so it cannot be fixed by an editor or a pre-commit hook (which we found sound for style-related rules).

Besides, we use Prettier for consistent styling which is an amazing tool, and we hope that they will someday include newline consistency in their formatting (it's only partially supported at the moment because it's technically difficult but they keep improving it at an impressive pace so who knows).

Would you be willing to disable this rule in the recommended ruleset?

JoshuaKGoldberg:
Absolutely 🙂. Would you like to send that PR?

If you want to make some improvements on the TSLint side, it'd be great if you could file issues there & link them back here. Eventually it seems reasonable to re-enable the rule once the flaws you mentioned are fixed.

JoshuaKGoldberg:
Fixed by #497 🚀

PR #497, opened 2018-09-03
Disable newline-before-return in recommended preset

astorije:
As discussed in #486.

Rather likely I forgot to change/add/remove something in the code. In particular, do I need to remove [this line](https://github.com/astorije/tslint-microsoft-contrib/blob/astorije/disable-newline-before-return/Gruntfile.js#L329)?


JoshuaKGoldberg:
This LGTM, but yes, I'm also suspicious about there being another place... but I can't find anything. Looks good so far - `grunt all` doesn't rewrite any files after this. Thanks!

PR #500, opened 2018-09-12
#395 Updated banned term rule to find banned terms in property names

### LINKED ISSUE #395: no-reserved-keywords not finding "catch" in all places
I have this code:
```ts
spyOn(customerService, "getCustomer").and.returnValue({
  then: (): {catch: () => void} => {
    return {
      catch: (): void => { return; }
    };
  }
});
```

Running with
- __TSLint version__: 5.7.0
- __TypeScript version__: 2.5.3
- __Running TSLint via__:  gulp

#### Actual behavior

It only reports a violation in the typedef

image

#### Expected behavior
I would expect it to also report a violation in the return statement below where the actual object property is named `catch`, in addition to the typedef above.

ghost:
closes #395

### DIFF
--- a/src/chaiPreferContainsToIndexOfRule.ts
+++ b/src/chaiPreferContainsToIndexOfRule.ts
@@ -15,6 +15,7 @@ export class Rule extends Lint.Rules.AbstractRule {
 
     public static metadata: ExtendedMetadata = {
         ruleName: 'chai-prefer-contains-to-index-of',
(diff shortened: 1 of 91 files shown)


JoshuaKGoldberg:
Well, looks like this from a deleted user - sorry to see you go! I'll go ahead and close this PR for housekeeping.

Ghost mystery friend, if you ever want to come back, I'd love to have this PR re-opened! 👻

PR #505, opened 2018-10-02
Fix #459: function-name rule should ignore computed names

### LINKED ISSUE #459: Method names should support symbol properties
The current function-name rule fails when implementing an iterable:
```typescript
const it: Iterable<any> = {
  // [tslint] Method name does not match /^[a-z][\w\d]+$/: [Symbol.iterator] (function-name)
  [Symbol.iterator](): any {}
}
```
It should not fail; symbol properties and the iterable protocol are standard parts of JS since many years ago and should be supported.

GabrielCastro:


JoshuaKGoldberg:
Thanks for the PR @GabrielCastro!

PR #507, opened 2018-10-02
Fix #437: Add undefined check before using toLowerCase

### LINKED ISSUE #437: react-a11y-image-button-has-alt throws on HTMLInputElement
This is a simple react component that consists of an `input` element only.

```js
import * as React from 'react'

interface IInputProps {
  type: 'text' | 'password'
}

class Input extends React.Component<IInputProps> {
  public render() {
    return <input type={this.props.type} />
  }
}

export default Input
```

`react-a11y-image-button-has-alt throws` throws this error:

```
The 'react-a11y-image-button-has-alt' rule threw an error in
'/mnt/app/src/components/Input/Input.tsx':

TypeError: Cannot read property 'toLowerCase' of undefined
    at ReactA11yImageButtonHasAltWalker.validateOpeningElement
    (/mnt/app/node_modules/tslint-microsoft-contrib/reactA11yImageButtonHasAltRule.js:66:77)
```

`tslint-microsoft-contrib` version is `5.0.3`

joostme:
Add `undefined` check before using `.toLowerCase()` on `getStringLiteral(typeAttribute)` to make the rule pass when the expression for the type of an input is `undefined`.

Fixes #437

### DIFF
--- a/src/reactA11yImageButtonHasAltRule.ts
+++ b/src/reactA11yImageButtonHasAltRule.ts
@@ -59,6 +59,7 @@ class ReactA11yImageButtonHasAltWalker extends Lint.RuleWalker {
         if (!typeAttribute
             || typeAttribute.initializer === undefined
             || !isStringLiteral(typeAttribute.initializer)
+            || getStringLiteral(typeAttribute) === undefined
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
Awesome, thanks @joostme!

PR #508, opened 2018-10-02
Allow same folder imports (fixes #493)

### LINKED ISSUE #493: Option request: allow same folder imports (no-relative-import)
Angular generates specs with same folder imports, like:

```ts
import { MyComponent } from './my.component';
```

So it would be really convenient if rule `no-relative-import` had an option to ignore imports that point to the same folder as the current file, with `./`.

With this option enabled, the `no-relative-import` rule would still have to report unnecessarily complex same folder imports like `./../my/my.component`.

IllusionMH:
Fixes #493

**Done:**
* Clean up for test file - fixed lint errors and removed unnecessary symbols.
* Added support for `allow-siblings` option.
* Updated rule metadata to describe new option.
* Updated README to mention new option.
* *(upd)* Rule is no longer marked as `typescriptOnly`
* checking for `..` deep inside import path left out of scope of this rule

**Questions:**

* at this moment rule allows paths like this `'absolute/path/…/path/file'` which works with TS and Webpack. Does it make sense to prevent this kind of paths in all cases?

* According to conversation in issue - arbitrary nesting is supported in relative paths. However it might lead for long relative import from files that are close to project root. May be it make sense to specify max allowed nesting for relative path? In our projects it is 2, if there is need for more - probably structure of components should be reconsidered.

    For example option might be converted to something like  `[true, {"allowed-relative-nesting-level": 2}]` this will allow `./file`, `./lvl1/file`, `./lvl1/lvl2/file`, but not `./lvl1/lvl2/lvl3/file`. `0` will allow only files in the same folder.

    What do you think?

**Other problems/questions:**

 * `package-lock.json` is updated "without reason". (Unable to install npm 5.7+ ATM to use `npm ci`) - updates not included in PR

* `recommended_ruleset.js` contains updates not related to updated rule - updates not included in PR

* Currently TS 3.0.1 is installed in this repo which leads to errors
    ```
    Running "ts:test-data" (ts) task
    Compiling...
    Cleared fast compile cache for target: test-data
    Using tsc v3.0.1
    node_modules/typescript/lib/lib.dom.d.ts(17113,18): error TS2451: Cannot redeclare block-scoped variable 'setImmediate'.
    node_modules/typescript/lib/lib.dom.d.ts(17114,18): error TS2451: Cannot redeclare block-scoped variable 'setImmediate'.
    test-data/NoStringBasedSetImmediateTestInput.ts(3,15): error TS2451: Cannot redeclare block-scoped variable 'setImmediate'.
    ```
    Commenting `declare const setImmediate: (...args: any[]) => number;` in later file helps.

    Probably TS version should be updated in `package.json`

* No lint for test files? Looks like there are a lot of violations (that are highlighted in editor), but no warnings/error with `grunt all`


JoshuaKGoldberg:
> paths like this `'absolute/path/…/path/file'`

If TS and Webpack allow them, it doesn't seem like it would make much sense to put extra effort in to block them. Redundant `..`s in paths could be a separate lint rule. +1 to how you have it now.

> Specify max allowed nesting for relative path

That's an interesting discussion. Would you mind holding off in this PR so we can discuss that in a separate issue filed once this goes in? Seems like that would be a bit trickier and it'd be nice to have this change in first.

> recommended_ruleset.js contains updates not related to updated rule - updates not included in PR

I can take care of those separately, thanks for the heads up

> Probably TS version should be updated in package.json

Indeed - I'll file a separate issue and push that through.

> No lint for test files

Seems like that should be a new issue as well, agreed.

IllusionMH:
@JoshuaKGoldberg thanks for review and comments.

>>Specify max allowed nesting for relative path
>
>That's an interesting discussion. Would you mind holding off in this PR so we can discuss that in a separate issue filed once this goes in? Seems like that would be a bit trickier and it'd be nice to have this change in first.

Yes, it can be discussed as a separate issue. I just wanted to know if it should be considered before merging this PR, because if  `allow-siblings` will be replaced by an object it might be a breaking change. Or it might be new rule `strict-import-paths` which will replacethis one.

> Redundant ..s in paths could be a separate lint rule

At this point there is asymmetry in implementation - `..` are not allowed at all when `allow-siblings` option is present (as was shown in initial issue), but can be used when no options passed (if not at the beginning of the sting).

Creating new rule will make `allow-sibling` options redundant, and users will have to decide between two conflicting rules `no-relative-imports` and `no-parent-imports` (suggested new rule) in their configuration which doesn't seems optimal.

Personally I lean towards forbidding `..` in both cases (simple code change). In this case it will still be single rule that controls import paths, even if will do a bit more that name suggests (may be renamed later to more speaking `strict-import-paths`) and might find few errors in existing codebases.

Also I forgot to note that this rule is no longer marked as `typescriptOnly` which might find new errors in existing codebases, but looks logical.

P.S. Sorry for new long post that introduces more questions than answers.

JoshuaKGoldberg:
> `..`

I'm a little lost... Clarifying for my own sake: if the idea for the `no-relative-imports` rule is to ban relative imports, IMO that's all it should do. Starting with `../` should always be banned by the rule and starting with `./` should be banned by the rule _if_ `allow-siblings` is not provided.

Whether there's a `../` somewhere deep within that folder path doesn't change the `no-relative-import` rule's logic, right? If that's correct, then this rule shouldn't check for that at all. Having a `../` deep within the path is just a quirk that could be managed elsewhere. If a new `strict-import-paths` rule gets added, that sounds fine, but for the scope of this rule it seems unrelated.

Am I missing something?

> no longer marked as `typescriptOnly`

That's fine - I'll mark this as a breaking change in the release notes. With services such as languages and linters, it's fairly impossible to avoid breaking changes in minor or even patch versions. This change certainly warrants at least a minor version update and not just a patch.

IllusionMH:
> Having a ../ deep within the path is just a quirk that could be managed elsewhere. If a new strict-import-paths rule gets added, that sounds fine, but for the scope of this rule it seems unrelated.

Agree. I've removed check for `../` deep inside path. Implementation and messages are only about start of the stings, while more strict checks will be considered for new rule (I think I will come up with finished rule proposal next week).

At this point I don't think that there are any open questions left for this rule and this PR can be merged if it passes review.

JoshuaKGoldberg:
Awesome, thanks so much @IllusionMH! I appreciate you walking me through this 😊.

One miscellaneous thought on the encompassing rule: I personally do the opposite of this by always using relative imports _except_ for aliased namespaces (e.g. using `"@project/bundle-name"` to refer to a Webpack bundle located at `/src/BundleName`)

PR #509, opened 2018-10-03
Update load-grunt-tasks package to 4.0.0

mateoholman:
While getting things setup to work on my first issue, I noticed that the current version of load-grunt-tasks requires a version of the minimatch package that has a RegEx DoS vulnerability. See:
https://medium.com/node-security/minimatch-redos-vulnerability-590da24e6d3c
https://www.npmjs.com/advisories/118
The recommended update is to load-grunt-tasks^4.0.0
That is a major version bump, but it does not seem to introduce any breaking changes. I'm totally new to the codebase, but thought it was worth mentioning! Also willing to open an issue if you would prefer to track it there.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2071,45 +2071,15 @@
       "dev": true
     },
     "load-grunt-tasks": {
-      "version": "3.2.0",
-      "resolved": "https://registry.npmjs.org/load-grunt-tasks/-/load-grunt-tasks-3.2.0.tgz",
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
Oh hey, thanks for bringing this up @mateoholman! I'll take a look now - an `npm audit` run following an [`ncu -a -u`](https://www.npmjs.com/package/npm-check-updates) would probably be good.

I'll close this pending a more deep set of updates, but really appreciate you pointing this out! 🙌

PR #520, opened 2018-10-04
Add rule: non-literal-fs-path

### LINKED ISSUE #253: new security rule: detect-non-literal-fs-filename
https://github.com/nodesecurity/eslint-plugin-security/blob/master/rules/detect-non-literal-fs-filename.js

medusalix:
Closes #253

Implements a rule that detects `fs` function calls with a non literal filepath.
This is my first time adding a new rule, please feel free to reach out if I've done anything incorrectly 😃.


src/tests/TestHelper.ts
```diff
@@ -128,6 +128,7 @@ export module TestHelper {
             program = ts.createProgram([inputFileOrScript], { });
         }
         if (inputFileOrScript.match(/.*\.ts(x)?$/)) {
+            /* tslint:disable:non-literal-fs-path */
```
JoshuaKGoldberg: Ah, `tslint:disable` disables the rule for the rest of the file. You're looking for:

```typescript
// tslint:disable-next-line non-literal-fs-path
```

medusalix: Oops, dind't know about that, should be fixed now.

ghs:
How would you deal with dynamic paths in such case? Easy would be to disable the rule if the code have a trust what and where from the file system will be used, but still..

Let say: fs.readdir('/uploads/' + year + '/' + month)...

JoshuaKGoldberg:
@ghs that's correct, you'd likely want to either disable this rule with a `// tslint:disable-next-line` style comment or in your `tslint.json`.

PR #523, opened 2018-10-06
Checks image alt for "image"... (fixes #219)

### LINKED ISSUE #219: Update react-a11y-img-has-alt - img alt prop must not contain the word "image", "picture", or "photo"
Enforce img alt attribute does not contain the word image, picture, or photo. Screenreaders already announce img elements as an image. There is no need to use words such as image, photo, and/or picture.

img alt prop must not contain the word "image", "picture", or "photo"
accept some parameters here and allow configuration
not case sensitive!

cthulhu-uu:
Looks for "image," "photo," and "picture" in img alt string and raises the added error.
Also matches words separated by underscore and dash. Fixes #219


src/reactA11yImgHasAltRule.ts
```diff
@@ -99,6 +104,8 @@ class ImgHasAltWalker extends Lint.RuleWalker {
                 getFailureStringNoAlt(tagName)
             );
         } else {
+            const altAttributeValue = getStringLiteral(altAttribute);
+            const hasImageInAlt: boolean = !!String(altAttributeValue).toLowerCase().match(/(\b|[_-])(photo|image|picture)(\b|[_-])/);
```
JoshuaKGoldberg: This is some fun JS syntax 🙂 but IMO the error message would look a little cleaner if it only said the redundant descriptor(s) actually in the tag. Instead of a single regexp, could you please have it collect the list and print only the relevant ones?

cthulhu-uu:
@JoshuaKGoldberg left the regexp in but made sure it lists which word was found. Let me know if this is ok!

cthulhu-uu:
haha sorry @JoshuaKGoldberg I'm still new to GitHub ❤️

src/reactA11yImgHasAltRule.ts
JoshuaKGoldberg: > really new to programming

...and you're working on an open source project that uses the TypeScript APIs to do static analysis? Great start! 👏

> is the single regex to complex in any form?

Sorry, I don't follow what you mean by that. Are you referring to just here, or regular expressions in general, or...?

JoshuaKGoldberg:
Just a heads up: an [unrelated PR](https://github.com/Microsoft/tslint-microsoft-contrib/pull/539) was merged in that touched pretty much every test file, so you'll probably have to change your unit tests slightly.

```typescript
name: 'file.ts',
```

...you'll want:

```typescript
Utils.absolutePath(file.ts')
```

JoshuaKGoldberg:
Ping @cthulhu-uu, are you still available to work on this PR? It'd be great to see it come in!

JoshuaKGoldberg:
Closing now to keep the queue small so others feel welcome to take this on themselves. Feel free to ping me if you'd like to pick this back up!

PR #524, opened 2018-10-06
Fix #394: Add ignore-case and ignore-whitespace options to react-a11y-anchors

### LINKED ISSUE #394: react-a11y-anchors where link text differs by case & whitespace
The rule "Links with the same HREF should have the same link text." in react-a11y-anchors is triggered when the links differ by case, as well as due to irrelevant white space (i.e. indentation). Could this be relaxed to allow differences in case, and to ignore leading and trailing whitespace?

nadyafebi:
Fixes #394.

Added toggleable `ignore-case` and `ignore-whitespace` options to `react-a11y-anchors` rule.

Also added unit tests and documentation for these options.


src/tests/ReactA11yAnchorsRuleTests.ts
```diff
@@ -259,6 +261,30 @@ describe('reactA11yAnchorsRule', () : void => {
         TestHelper.assertViolations(ruleName, script, [ ]);
     });
 
+    it('should pass when identical hrefs have texts with different cases on ignore-case enabled', () : void => {
```
JoshuaKGoldberg: Good unit tests for the case of the rules enabled! But: could you please also add equivalent tests for the same cases but with your new options not enabled? That way we still capture the case of the rule accidentally always enabling those options.

src/reactA11yAnchorsRule.ts
```diff
@@ -95,6 +117,29 @@ class ReactA11yAnchorsRuleWalker extends ErrorTolerantWalker {
...
+            text2 = text2.toLowerCase();
+            altText1 = altText1.toLowerCase();
+            altText2 = altText2.toLowerCase();
+        }
+
+        if (this.ignoreWhitespace) {
+            text1 = text1.trim();
+            text2 = text2.trim();
+            altText1 = altText1.trim();
+            altText2 = altText2.trim();
```
JoshuaKGoldberg: Is it intentional that this only trims instead of replacing whitespace? Should `"a b"` still be considered different from `"ab"`?

I would have thought the rule should remove all whitespace? Perhaps not a blocker for this PR, but maybe the `ignore-whitespace` should be allowed to be `"all" | "trim"` to allow either behavior?

src/tests/ReactA11yAnchorsRuleTests.ts
nadyafebi: Good idea! I've added the new unit tests on the latest commit.

src/reactA11yAnchorsRule.ts
nadyafebi: The original issue is only concerned about leading/trailing whitespace, but I agree that there should be an option to enable either behavior. I will make some changes then.

nadyafebi:
Hi @JoshuaKGoldberg, I've updated the pull request with the proposed changes. So now to use the new options, user can just update their `tslint.json` with this:
```
react-a11y-anchors: {
    true,
    "ignore-case",
    {
        "ignore-whitespace": "all"
    }
}

JoshuaKGoldberg:
The Node 6 build timed out on an unrelated rule... re-running. Pretty sure it was just intermittent noise. https://travis-ci.org/Microsoft/tslint-microsoft-contrib/jobs/438391623

PR #528, opened 2018-10-08
Add launch.json for vscode

roblourens:
Can I interest you in a launch config for vscode? I've found this useful and I see that you already have some `.vscode/` config. (But it won't hurt my feelings if you don't take it.)


JoshuaKGoldberg:
Whoohoo!

Issue #527, opened 2018-10-08
Use (j|t)sx in getLanguageVariant, rather than just tsx


JoshuaKGoldberg:
...and of course, immediately after #525 was resolved, I remembered that we haven't updated for TypeScript being able to run on `.jsx` files as well as `.tsx`.

dnguneratne:
Can I work on this?

JoshuaKGoldberg:
Absolutely @dnguneratne, please do! It should be a quick change to the regular expression mentioned in that issue.

dnguneratne:
I think I need a bit of clarity here. 
In PR #525,

-- `src/utils/AstUtils.ts` --
`if (/.*\.tsx/i.test(node.fileName)) {` was replaced with 
`if (node.fileName.endsWith('.tsx')) {`.

If I am understanding this one correctly I should change
`if (node.fileName.endsWith('.tsx')) {` to
`if (node.fileName.endsWith('.(j|t)sx')) {` right?

JoshuaKGoldberg:
Hey @dnguneratne - looks like this was already resolved. All is well!

By `(j|t)` I was referring to regular expression syntax implying it could be satisfied by `.jsx` or `.tsx` at the end.

dnguneratne:
Got it. Thanks for the clarification.

PR #530, opened 2018-10-08
AstUtils.getLanguageVariant() now accepts .jsx files and ignores case

shutdown27:
I added the .jsx file extension to AstUtils.getLanguageVariant() as requested in issue #527
I also made it ignore the case of the extensions,
i think this behaviour was accidentaly removed in PR #526


JoshuaKGoldberg:
Super, thanks @shutdown27!

PR #534, opened 2018-10-09
feat(ReactThisBindingIssueRule): allow decorators to bind

### LINKED ISSUE #392: Add detection for @bind decorators in react-this-binding
It's common practice to do something like the following:

```tsx
import { bind } from 'decko';

class Foo extends Component {
  public render() {
    return <button onclick={this.sayHi}>Greet</button>;
  }

  @bind
  public sayHi() {
    alert('hi');
  }
}
```

Unfortunately the `react-this-binding-issue` rule isn't aware of this and gets tripped up.

I propose adding an option like `{ "bindDecorators": ["bind"] }` which looks at decorator names and flags functions with those decorators as having been bound. It can have sensible defaults like `["bind", "Bind"]`.

cyberhck:
closes #392


src/reactThisBindingIssueRule.ts
```diff
@@ -86,11 +104,29 @@ class ReactThisBindingIssueRuleWalker extends ErrorTolerantWalker {
...
         this.scope = new Scope(null);
         super.visitMethodDeclaration(node);
         this.scope = null;
     }
 
+    private isMethodBoundWithDecorators(node: ts.MethodDeclaration): boolean {
+        if (!(this.allowedDecorators.length > 0 && node.decorators && node.decorators.length > 0)) {
+            return false;
+        }
+        const bindingDecorators = node.decorators.find((decorator) => {
```
IllusionMH: `some` can be used instead of `find`.
Then conversion to `boolean` can be omitted later.

src/reactThisBindingIssueRule.ts
```diff
@@ -54,6 +69,9 @@ class ReactThisBindingIssueRuleWalker extends ErrorTolerantWalker {
             if (typeof(opt) === 'object') {
                 this.allowAnonymousListeners = opt['allow-anonymous-listeners'] === true;
             }
+            if (typeof opt === 'string') {
```
IllusionMH: IMHO, it would be better to add "speaking" property to object, instead of listing all passes strings as decorators. So setup will be like
```
"react-this-binding-issue": [true, {
    "allow-anonymous-listeners": true,
    "bind-decorators": ["autobind", "custom-bind"]
}]
```
instead of
```
"react-this-binding-issue": [true, "autobind", "customBind", { "allow-anonymous-listeners": true}]
```

But it's up to maintainers.

src/reactThisBindingIssueRule.ts
```diff
@@ -86,11 +104,29 @@ class ReactThisBindingIssueRuleWalker extends ErrorTolerantWalker {
...
         // in scope so we can make sure new function instances are not passed as JSX attributes
+        if (this.isMethodBoundWithDecorators(node)) {
+            this.boundListeners = this.boundListeners.concat('this.' + node.name.getText());
+        }
         this.scope = new Scope(null);
         super.visitMethodDeclaration(node);
         this.scope = null;
     }
 
+    private isMethodBoundWithDecorators(node: ts.MethodDeclaration): boolean {
```
cyberhck: instead of using `this.allowedDecorators`, if we pass it to this function, then it'd become pure, will change.

src/reactThisBindingIssueRule.ts
cyberhck: This is something I didn't find much examples of or documentation, how do I write a schema? because options is of type `any`

src/reactThisBindingIssueRule.ts
```diff
@@ -68,9 +69,9 @@ class ReactThisBindingIssueRuleWalker extends ErrorTolerantWalker {
         this.getOptions().forEach((opt: any) => {
             if (typeof(opt) === 'object') {
                 this.allowAnonymousListeners = opt['allow-anonymous-listeners'] === true;
-            }
-            if (typeof opt === 'string') {
-                this.allowedDecorators = this.allowedDecorators.concat(opt);
+                if (opt['bind-decorators']) {
```
IllusionMH: I think it might be safer to check it's an array before assigning it to `allowedDecorators`. Maybe even filter to string only.

cyberhck:
@IllusionMH can you verify if everything is correct? Specially the schema part :)

src/reactThisBindingIssueRule.ts
```diff
@@ -53,6 +71,14 @@ class ReactThisBindingIssueRuleWalker extends ErrorTolerantWalker {
         this.getOptions().forEach((opt: any) => {
             if (typeof(opt) === 'object') {
                 this.allowAnonymousListeners = opt['allow-anonymous-listeners'] === true;
+                if (opt['bind-decorators']) {
+                    const allowedDecorators: any[] = opt['bind-decorators'];
+                    if (allowedDecorators.constructor !== Array && allowedDecorators.some((decorator) => typeof decorator !== 'string')) {
```
IllusionMH: `Array.isArray` should be more readable/convenien.
`||` makes more sense here. Because second check should be performed only if `opt['bind-decorators']` is an array (otherwise it will be runtime error)

test-data/ReactThisBinding/ReactThisBindingIssueWithDecorator-passing.tsx
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    // this listener is a lambda, so it has its 'this' reference bound correctly.
+    private listener3: () => void = (): void => {};
+
+    public render() {
+        return <div
+            onClick={this.listener1}>
+            onMouseOver={this.listener2}>
+            onMouseDown={this.listener3}>
```
JoshuaKGoldberg: You've got a few syntax errors here in the test: `>` should only come after the last attribute.

cyberhck: Ahh, my bad, I copied from existing one and I didn't notice that for some reason :smile: , I'll fix the original file as well :)

cyberhck:
I just rebased, merged, can you please check? I did sign up for hacktoberfest before starting on this :)

src/reactThisBindingIssueRule.ts
```diff
@@ -53,6 +71,17 @@ class ReactThisBindingIssueRuleWalker extends ErrorTolerantWalker {
         this.getOptions().forEach((opt: any) => {
             if (typeof(opt) === 'object') {
                 this.allowAnonymousListeners = opt['allow-anonymous-listeners'] === true;
+                if (opt['bind-decorators']) {
+                    const allowedDecorators: any[] = opt['bind-decorators'];
+                    if (
+                        !Array.isArray(allowedDecorators.constructor)
```
IllusionMH: Check should be for `allowedDecorators` not it's constructor.

Error below can be rephrase to mention that `bind-decorators` should be a list of strings, because now it will fail if string is passed, and not an array

cyberhck: ahh, I missed the allowedDecorators.constructor part for some reason, my bad, will fix after I wake up now.

IllusionMH:
👍 
No more questions/comments from my side.

cyberhck:
Hey maintainers, is there anything I can do to make it better? or get merged? I kinda want this :)

JoshuaKGoldberg:
Heya @cyberhck sorry for the delay! Looking now.

cyberhck:
No problem, thanks for merging :) (a t-shirt awaits)

cyberhck:
I can't believe we missed this: #813

PR #538, opened 2018-10-09
Fix #433: react-a11y-anchors should accept child component

### LINKED ISSUE #433: react-a11y-anchors should accept content when given a child component
`react-a11y-anchors` passes on the following snippets (assuming each line is a different file as I am focusing this issue on text and not `href`):

```tsx
<a href="https://example.com">Hello world</a>
<a href="https://example.com"><TextWrapper>Hello world</TextWrapper></a>
```

However, it fails on the following ones:

```tsx
<a href="https://example.com"><TextWrapper /></a>
<a href="https://example.com"><TextWrapper value="Hello world" /></a>
```

It's not an uncommon situation, for example when rendering translated strings.

I think it would be reasonable to take [`eslint-plugin-jsx-a11y`'s approach](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-has-content.md): accept any React component as valid content, unless it has a `aria-hidden` prop.
Thoughts?

br1anchen:
Fixes #433

### DIFF
--- a/src/reactA11yAnchorsRule.ts
+++ b/src/reactA11yAnchorsRule.ts
@@ -29,6 +29,8 @@ export const SAME_HREF_SAME_TEXT_FAILURE_STRING: string =
     'Links with the same HREF should have the same link text.';
 export const DIFFERENT_HREF_DIFFERENT_TEXT_FAILURE_STRING: string =
     'Links that point to different HREFs should have different link text.';
+export const ACCESSIBLE_HIDDEN_CONTENT_FAILURE_STRING: string =
+    'Link content can not be hidden for screen-readers by using aria-hidden attribute.';
 
 /**
  * Implementation of the react-a11y-anchors rule.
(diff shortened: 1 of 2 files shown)


src/reactA11yAnchorsRule.ts
```diff
@@ -245,6 +252,11 @@ class ReactA11yAnchorsRuleWalker extends ErrorTolerantWalker {
         } else if (root.kind === ts.SyntaxKind.JsxExpression) {
             const expression: ts.JsxExpression = <ts.JsxExpression>root;
             title += this.anchorText(expression.expression);
+        } else if (isChild && root.kind === ts.SyntaxKind.JsxSelfClosingElement) {
+            const jsxSelfClosingElement = <ts.JsxSelfClosingElement>root;
+            if (jsxSelfClosingElement.tagName.getText() !== 'img') {
```
br1anchen: Did not find a good way to diff `html` jsx element than customised jsx element in this case.
So currently only check if children component is not img since `imageAlt` will check `altText ` with `anchorText `.
Passes all current test cases though.

JoshuaKGoldberg: Yeah this seems fine for now. I don't know a better way than to directly get the text of the `tagName`...

PR #539, opened 2018-10-09
Handle relative paths in exceptions for react-no-dangerous-html rule

### LINKED ISSUE #537: Suppressions for react-no-dangerous-html require absolute path
Suppressions for react-no-dangerous-html only work when the file attribute contains the absolute path to the file.

I believe the problem is line 120 in `src/reactNoDangerousHtmlRule.ts` when the file attribute is compared for equality to a `SourceFile.fileName` which is an absolute path.

kevinwilde:
Fixes #537

### DIFF
--- a/src/tests/TestHelper.ts
+++ b/src/tests/TestHelper.ts
@@ -1,7 +1,7 @@
 import * as Lint from 'tslint';
 import * as fs from 'fs';
 import * as chai from 'chai';
-import * as path from 'path';
+import {Utils} from '../utils/Utils';
 import * as ts from 'typescript';
 import {ErrorTolerantWalker} from '../utils/ErrorTolerantWalker';
 
@@ -137,9 +137,9 @@ export module TestHelper {
         } else {
             let filename: string;
             if (inputFileOrScript.indexOf('import React') > -1) {
-                filename = path.resolve('file.tsx');
+                filename = Utils.absolutePath('file.tsx');
             } else {
-                filename = path.resolve('file.ts');
+                filename = Utils.absolutePath('file.ts');
             }
 
             const linter = new Lint.Linter(options, useTypeChecker ? program : undefined);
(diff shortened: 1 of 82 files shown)


src/tests/TestHelper.ts
```diff
@@ -136,9 +137,9 @@ export module TestHelper {
         } else {
             let filename: string;
             if (inputFileOrScript.indexOf('import React') > -1) {
-                filename = 'file.tsx';
+                filename = path.resolve('file.tsx');
             } else {
-                filename = 'file.ts';
+                filename = path.resolve('file.ts');
```
kevinwilde: The tests previously just used 'file.ts(x)' as a mock filename. Since `reactNoDangerousHtmlRule` now compares `path.resolve(exception.file)` (which returns an absolute path) to `this.getSourceFile().fileName` (which also returns an absolute path when running tslint on an actual project) we need the tests to provide an absolute path for the filename, as well.

Unfortunately, this means I had to change a lot of test files, so this PR has a lot of line changes.

kevinwilde:
Ah shoot. I guess `SourceFile.fileName` uses `/` on Windows, but `path.resolve(..)` uses `\\`. Need to fix the tests

src/tests/TestHelper.ts
```diff
@@ -136,9 +137,9 @@ export module TestHelper {
         } else {
             let filename: string;
             if (inputFileOrScript.indexOf('import React') > -1) {
-                filename = 'file.tsx';
+                filename = Utils.absolutePath('file.tsx');
             } else {
-                filename = 'file.ts';
+                filename = Utils.absolutePath('file.ts');
```
kevinwilde: The tests previously just used 'file.ts(x)' as a mock filename. Since `reactNoDangerousHtmlRule` now compares `Utils.absolutePath(exception.file)` to `this.getSourceFile().fileName` (which also returns an absolute path when running tslint on an actual project) we need the tests to provide an absolute path for the filename, as well.

Unfortunately, this means I had to change a lot of test files, so this PR has a lot of line changes.

PR #542, opened 2018-10-12
Fix #362: Add an `allow-for-loop` option to `no-increment-decrement`

### LINKED ISSUE #362: Allow increment in for loop statements
Code like this is perfectly idiomatic and has little risk of causing an issue:

```
for (let i = 0; i < length; i++) {
  // ...
}
```

It would be nice to have an option to allow increment operators in this context.

Retsam:
Fixes #362. If the `allow-for-loop` option is passed to `no-increment-decrement`, then increment statements can be used in the headers of for loops. 

I wasn't sure how fine-grained we wanted to be: this allows increment and decrement operators anywhere in the for loop header; but perhaps they should only be allowed in the "incrementor" part of the header.


JoshuaKGoldberg:
Just a heads up @Retsam: an [unrelated PR](https://github.com/Microsoft/tslint-microsoft-contrib/pull/539) was merged in that touched pretty much every test file, so you'll probably have to change your unit tests slightly.

```typescript
name: 'file.ts',
```

...you'll want:

```typescript
Utils.absolutePath(file.ts')
```

Retsam:
@JoshuaKGoldberg Thanks for the heads-up.

JoshuaKGoldberg:
Thanks for this @Retsam! The change is good so I took the liberty of merging in `master`.

PR #543, opened 2018-10-12
#278 Implement react-a11y-no-onchange.

### LINKED ISSUE #278: new a11y rule: no-onchange
no-onchange

Enforce usage of onBlur over onChange on select menus for accessibility.

Also, OnChange event handler should not be used with the select element.
    • http://oaa-accessibility.org/wcag20/rule/62/

sahalsaad:
Fixes #278 
- OnChange event handler should not be used with the select element.


src/reactA11yNoOnchangeRule.ts
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Implementation of the react-a11y-no-onchange rule.
+ */
+export class Rule extends Lint.Rules.AbstractRule {
+
+    public static metadata: ExtendedMetadata = {
+        ruleName: 'react-a11y-no-onchange',
+        type: 'functionality',
+        description: 'For accessibility of select menus, onChange event handler should not be used with the select element.',
```
JoshuaKGoldberg: For users who are new to accessibility concerns, it's confusing why the `onChange` event handler isn't allowed with the select element. Could you please add a `rationale` mentioning the preference to use `onBlur`?

src/reactA11yNoOnchangeRule.ts
```diff
@@ -0,0 +1,64 @@
...
+        options: 'string[]',
+        optionsDescription: 'Additional tag name to validate.',
```
JoshuaKGoldberg: While you're in the area, would you mind slightly changing this to indicate that any number of tag name**s** are allowed? It would be easy to assume only one tag name is allowed from just seeing this description.

sahalsaad:
Updated my PR. Let me know if you have a better description. I can update it.

JoshuaKGoldberg:
> why the test failed

An [unrelated PR](https://github.com/Microsoft/tslint-microsoft-contrib/pull/539) was merged in that touched pretty much every test file, so you'll have to change your unit tests slightly.

```typescript
name: 'file.ts',
```

...you'll want:

```typescript
Utils.absolutePath('file.ts')
```

Issue #535, opened 2018-10-09
ReactThisBindingIssueRule should use Sets on the inside


JoshuaKGoldberg:
https://node.green/#ES2015-built-ins-Set

`Set`s have been supported in Node for quite some time. It'd be a bit more clear and efficient on super long files to use them instead of arrays for the internal members.

Not a huge priority (functionality shouldn't change), and should come after #534 is merged.

cyberhck:
Do you mean for private properties? Where we store the data?
Is that supposed to contain everything? If so, how about the data types? Will there be no types? Or we create separate sets for separate type of data? I want to take a stab on this issue after that's merged :slightly_smiling_face:

PR #544, opened 2018-10-12
#535 In ReactThisBindingIssueRule internal arrays are replaced by sets

jonascarlsonschmidt:
Implemented #535 

- All internal arrays are replaces by set


src/reactThisBindingIssueRule.ts
```diff
@@ -243,8 +243,8 @@ class ReactThisBindingIssueRuleWalker extends ErrorTolerantWalker {
 
                                         const rightPropText = AstUtils.getFunctionTarget(callExpression);
                                         if (leftPropText === rightPropText) {
-                                            if (result.indexOf(rightPropText) === -1) {
-                                                result.push(rightPropText);
+                                            if (!result.has(rightPropText)) {
```
JoshuaKGoldberg: Fun fact: you can actually just skip this `.has`, since a `.add` on something that already exists in the Set does nothing.

PR #547, opened 2018-10-13
Fix #275 Add new rule for accessible input elements

### LINKED ISSUE #275: new a11y rule: accessible input elements
Edit boxes and text areas must include default, place-holding characters

https://www.w3.org/TR/WCAG10/wai-pageauth.html#tech-place-holders
https://www.w3.org/TR/WCAG10-HTML-TECHS/#forms-specific

br1anchen:
fixes #275


src/reactA11yInputElementsRule.ts
```diff
@@ -0,0 +1,70 @@
...
+            if (isEmpty(attributes.placeholder)) {
+                this.addFailureAt(node.getStart(), node.getWidth(), MISSING_PLACEHOLDER_TEXTAREA_FAILURE_STRING);
+            }
+        }
+        super.visitJsxSelfClosingElement(node);
+    }
+
+    protected visitJsxElement(node: ts.JsxElement): void {
+        const tagName = node.openingElement.tagName.getText();
+        const attributes: { [propName: string]: ts.JsxAttribute } = getJsxAttributesFromJsxElement(node);
```
JoshuaKGoldberg: One small performance concern: for both `attributes` variables, could you please only call `getJsxAttributesFromJsxElement` if the `tagName` is one we care about? For files with many large JSX elements, this'll help avoid unnecessary processing time in the rule collecting attributes on other elements.

tslint.json
```diff
@@ -3,7 +3,10 @@
     "dist/src"
```
JoshuaKGoldberg: Looks like your auto-formatter got to `tslint.json` 😉 please revert these changes.

src/reactA11yInputElementsRule.ts
br1anchen: Good point. Will change.

PR #554, opened 2018-10-14
ensure map areas have non-empty alts

### LINKED ISSUE #223: Update react-a11y-img-has-alt - area elements of image maps have alternate text
area elements of image maps have alternate text - wcag2a, wcag111, section508, section508.22.a - From area-alt

Chovin:
closes #223


JoshuaKGoldberg:
> if they're both there together in their "literal" form

Sorry @Chovin I don't follow. Do you mean that the only way to tell is if the literals are together, like this?

```tsx
<map ...>
    <area ... />
</map>
```

...rather than:

```tsx
<SomeComponentThatReturnsAMap>
    <area ... />
</SomeComponentThatReturnsAMap>
```

If that's the case, then yes - there isn't a _(known?)_ good way to validate what JSX elements are returned by a method.

Chovin:
@JoshuaKGoldberg ah I see. I'm assuming that's the best we can do here and just ignore cases like this?

```ts
<SomeComponentThatReturnsAMap>
    <area ... />
</SomeComponentThatReturnsAMap>
```

Chovin:
hmm. `parent.type === 'map'` ?
Sorry, this is my first time touching React 🙃

Chovin:
Awesome! I’ll get to that test soon. Thanks!

JoshuaKGoldberg:
Ping @Chovin, are you still available to finish up this PR? Looks like it'd be a good one to have in.

Chovin:
I’d like to get back to this but will have to do so in a few weeks. Thanks for following up @JoshuaKGoldberg !

JoshuaKGoldberg:
Closing now to keep the queue small so others feel welcome to take this on themselves. Feel free to ping me if you'd like to pick this back up @Chovin!

Issue #400, opened 2017-10-30
Again audit rules that have already been added in TSLint


JoshuaKGoldberg:
If they're in TSLint proper, remove them from the README.md & give a warning in the file for a version or two instead of checking the rule (before removing the file)?

JoshuaKGoldberg:
Scrolling through file lists and READMEs...

Rules to remove:
* `function-name`: superseded by `tslint`'s `variable-name`
* `missing-jsdoc`: superseded by `tslint`'s `file-header`
* `no-duplicate-switch-case`: superseded by `tslint`'s `no-duplicate-switch-case`
* `no-empty-interfaces`: superseded by `tslint`'s `no-empty-interface`
* `no-reserved-keywords`: superseded by `tslint`'s `no-reserved-keywords`
* `no-unnecessary-field-initialization`: superseded by `tslint`'s `no-unnecessary-initializer`
* `prefer-array-literal`: superseded by `tslint`'s `array-type`
* `prefer-type-cast`: superseded by `tslint`'s `no-angle-bracket-type-assertion`
* `promise-must-complete`: superseded by `tslint`'s `no-floating-promises`
* `react-tsx-curly-spacing`: superseded by `tslint-react`'s `jsx-curly-spacing`
* `valid-typeof`: superseded by `tslint`'s `typeof-compare`, which was then superseded by the TS compiler

PR #559, opened 2018-10-15
Marked missing-jsdoc as deprecated

JoshuaKGoldberg:
Continues #400.


JoshuaKGoldberg:
Ehh, will just do #562 as a superset of this.

PR #565, opened 2018-10-15
Used undefined instead of null.

### LINKED ISSUE #490: Stop using null unnecessarily in source code
There are many places in source code that use `null` and `undefined` interchangeably. As a code cleanup task, it'd be nice to consolidate on only using `undefined` whenever possible. This is what most libraries do, including TSLint and TypeScript.

reduckted:
Fixes #490.

I've changed the code to use `undefined` instead of `null`, except where it's actually needed, which was only a handful of places:

* Creating an object without a prototype [here](https://github.com/Microsoft/tslint-microsoft-contrib/blob/dea2835101911f669dde46387833a8e57e054e89/src/noUnexternalizedStringsRule.ts#L51-L52)
* Checking the results of regular expression tests [here](https://github.com/Microsoft/tslint-microsoft-contrib/blob/dea2835101911f669dde46387833a8e57e054e89/src/maxFuncBodyLengthRule.ts#L133), [here](https://github.com/Microsoft/tslint-microsoft-contrib/blob/dea2835101911f669dde46387833a8e57e054e89/src/noJqueryRawElementsRule.ts#L73) and [here](https://github.com/Microsoft/tslint-microsoft-contrib/blob/dea2835101911f669dde46387833a8e57e054e89/src/noRegexSpacesRule.ts#L37)

There's one thing I'm not sure if I should change (I've left it unchanged for now). The [additional_rule_metadata.json](https://github.com/Microsoft/tslint-microsoft-contrib/blob/dea2835101911f669dde46387833a8e57e054e89/additional_rule_metadata.json#L347-L355) and [recommended_ruleset.js](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/recommended_ruleset.js#L141) files both suggest turning the `no-null-keyword` off. Given that it's now turned _on_, do those files need to change?

### DIFF
--- a/src/chaiPreferContainsToIndexOfRule.ts
+++ b/src/chaiPreferContainsToIndexOfRule.ts
@@ -17,7 +17,7 @@ export class Rule extends Lint.Rules.AbstractRule {
         ruleName: 'chai-prefer-contains-to-index-of',
         type: 'maintainability',
         description: 'Avoid Chai assertions that invoke indexOf and compare for a -1 result.',
-        options: undefined,
+        options: null, // tslint:disable-line:no-null-keyword
         optionsDescription: '',
         typescriptOnly: true,
         issueClass: 'Non-SDL',
--- a/src/chaiPreferContainsToIndexOfRule.ts
+++ b/src/chaiPreferContainsToIndexOfRule.ts
@@ -17,7 +17,7 @@ export class Rule extends Lint.Rules.AbstractRule {
         ruleName: 'chai-prefer-contains-to-index-of',
         type: 'maintainability',
         description: 'Avoid Chai assertions that invoke indexOf and compare for a -1 result.',
-        options: null,
+        options: undefined,
         optionsDescription: '',
         typescriptOnly: true,
         issueClass: 'Non-SDL',
@@ -51,7 +51,7 @@ class ChaiPreferContainsToIndexOfRuleWalker extends ErrorTolerantWalker {
     }
 
     private isFirstArgumentNegative1(node: ts.CallExpression): boolean {
-        if (node.arguments != null && node.arguments.length > 0) {
+        if (node.arguments !== undefined && node.arguments.length > 0) {
             const firstArgument: ts.Expression = node.arguments[0];
             if (firstArgument.getText() === '-1') {
                 return true;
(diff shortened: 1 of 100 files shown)


src/chaiPreferContainsToIndexOfRule.ts
```diff
@@ -17,7 +17,7 @@ export class Rule extends Lint.Rules.AbstractRule {
         ruleName: 'chai-prefer-contains-to-index-of',
         type: 'maintainability',
         description: 'Avoid Chai assertions that invoke indexOf and compare for a -1 result.',
-        options: null,
+        options: undefined,
```
JoshuaKGoldberg: This one always confused me. TSLint itself describes the rule's options as:

```typescript
Schema of the options the rule accepts.
The first boolean for whether the rule is enabled or not is already implied.
This field describes the options after that boolean.
If null, this rule has no options and is not configurable.
```

https://github.com/palantir/tslint/blob/237fd905a9fb3c9ca9325d2da6e8145e4a44fe51/src/language/rule/rule.ts#L64

...so we should probably default this to `null`.

JoshuaKGoldberg:
> The additional_rule_metadata.json and recommended_ruleset.js files both suggest turning the no-null-keyword off

Yes, let's turn them back on there. Those files are a little older.

src/chaiPreferContainsToIndexOfRule.ts
reduckted: Ah, yes, you're absolutely right. I had a look at that comment as well as how the options are used in TSLint before I made the changes and it looked like it shouldn't matter if undefined is used, but of course the _one_ place I didn't look at was what TSLint uses for their own rules, and they use null.

astorije:
@JoshuaKGoldberg, I'm having a lot of issues with this change in the recommended ruleset.

As much as I would love to stick to `undefined` in favor of `null` in our own codebase, the reality is that a lot of (popular) libraries out there do not support this.
You noted that TSLint itself has a different behavior on `null`/`undefined`, it's also true for `react-apollo` (and other Apollo-related libraries like `apollo-cache-inmemory`), and I'm sure I'll discover plenty more as I'm updating this plugin in our repos.
I'm also using https://github.com/godmodelabs/flora-sql-parser which strives on `null` values. While it's a small library and I could disable this rule on a per-file basis, it is built upon https://github.com/pegjs/pegjs which is much more widely used, and generates a ton of `null`s everywhere.

Are you at all opposed to removing this from the recommended ruleset?

reduckted:
> @astorije I could disable this rule on a per-file basis...

You can turn the rule off in the config file and still use the recommended ruleset.

```json
{
    "extends": ["tslint-microsoft-contrib"],
    "rulesDirectory": ["node_modules/tslint-microsoft-contrib"],
    "rules": {
        "no-null-keyword": false
    }
}
```

As the readme says,

>  To start, you can enable our recommended defaults...You can then disable rules you don't find useful.

astorije:
@reduckted, thanks, we're doing that for a few rules indeed.

FWIW, I do find this rule useful, I just don't think it should be recommended/enabled by default. Many packages out there do not play nice with this, and as we noticed on our end, fixing violations recommended by this rule on things outside of our control (dependencies) will often break.

PR #570, opened 2018-10-20
Enabled the no-any rule.

### LINKED ISSUE #566: Enable no-any in source code
Spin-off from #491. We should:

1. Enable the rule in `tslint.json` and `recommended_ruleset.js`
2. Fix up any violations of the rule in source code

reduckted:
Fixes #566.

There were no major problems making these changes, but there are a few things that I'll point out in comments against the changes.

### DIFF
--- a/src/exportNameRule.ts
+++ b/src/exportNameRule.ts
@@ -64,9 +64,7 @@ export class Rule extends Lint.Rules.AbstractRule {
             return options.ruleArguments[0];
         }
         if (options instanceof Array) {
-            return typeof options[0] === 'object' ?
-                options[0].allow :
-                <string[]><any>options; // MSE version of tslint somehow requires this
+            return typeof options[0] === 'object' ? options[0].allow : options;
         }
         return undefined;
     }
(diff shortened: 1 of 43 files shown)


src/tests/FixNoRequireImportsFormatterTests.ts
```diff
@@ -40,7 +40,7 @@ import TestHelper = require('./TestHelper');
 `;
 
         const formatter = new FormatterForTesting(input);
-        formatter.format(TestHelper.runRule(ruleName, null, input).failures);
+        formatter.format(TestHelper.runRule(ruleName, undefined, input).failures);
```
reduckted: These were somehow missed when `no-null-keyword` was turned on (seems like the grunt task isn't linting everything it should), but the type of `userOptions` in the `runRule()` signature changed, so these had to be fixed up anyway.

src/tests/ImportNameRuleTests.ts
```diff
@@ -146,7 +146,7 @@ describe('importNameRule', () : void => {
             import baseChartOptions = require('common/component/chart/options/BaseChartOptions');
         `;
 
-        const options: any[] = [ true, {
+        const options = [ true, {
```
reduckted: I cheated slightly here and let the type of `options` be inferred. It sort of seems like it goes against the conventions, but it didn't break any lint rules. 🤷‍♂️

src/utils/AstUtils.ts
```diff
@@ -16,11 +17,13 @@ export module AstUtils {
 
     export function getFunctionName(node: ts.CallExpression | ts.NewExpression): string {
         const expression: ts.Expression = node.expression;
-        let functionName: string = (<any>expression).text;
-        if (functionName === undefined && (<any>expression).name) {
-            functionName = (<any>expression).name.text;
+        if ('text' in expression) {
```
reduckted: This was an interesting one. Just using `expression.name.getText()` caused a whole bunch of tests to fail, so this does need to look for a `text` property on the expression, but it can't look for a `getText()` method, because that caused a bunch of different tests to fail.

src/utils/AstUtils.ts
```diff
@@ -110,44 +113,24 @@ export module AstUtils {
         /* tslint:enable:no-console */
     }
 
-    export function isPrivate(node: ts.Node): boolean {
-        /* tslint:disable:no-bitwise */
-        if ((<any>ts).NodeFlags.Private !== undefined) {
```
reduckted: The modifier flag types were moved out of the `NodeFlags` enum just before TypeScript 2.1. Since the minimum version of TypeScript specified in the peer dependencies here is 2.1.0, I figured it was time to remove this and just look at the `ModifierFlags`.

src/exportNameRule.ts
```diff
@@ -64,9 +64,7 @@ export class Rule extends Lint.Rules.AbstractRule {
             return options.ruleArguments[0];
         }
         if (options instanceof Array) {
-            return typeof options[0] === 'object' ?
-                options[0].allow :
-                <string[]><any>options; // MSE version of tslint somehow requires this
```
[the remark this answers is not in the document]
JoshuaKGoldberg: > `MSE version of tslint`

I don't know what MSE is but agreed that it's probably outdated. *M*icro*s*oft... *E*nterprise? 

/cc @HamletDRC in case this is relevant?

JoshuaKGoldberg:
Coming back to this @reduckted, I think I was wrong to ask to use `options: null`. I haven't found any problems with using `options: undefined` in TSLint _(why would there be? it's just for docs generation)_ and there's no motivation to use `null`. Sorry for the extra churn 😊 I'll switch it back to the superior `undefined`.

PR #574, opened 2018-10-23
Fix #571: Avoid redundant rel values in react-anchor-blank-noopener

### LINKED ISSUE #571: target="_blank" rel="noreferrer" implies noopener
```html
<a href="#" target="_blank" rel="noreferrer"></a>
```

> [tslint] Anchor tags with target="_blank" should also include rel="noopener noreferrer"

But, as pointed out here: https://twitter.com/_developit/status/1054422343609655296

> `<a rel="noreferrer">` implies noopener. Specifying both is redundant.

https://github.com/yannickcr/eslint-plugin-react/issues/2022

esuau:
## Problem

Fixes #571
The rule `react-anchor-blank-noopener` was designed to avoid security vulnerabilities by the use of `rel="noopener noreferrer"`  in anchor tags containing `target="_blank"`. It appears that `rel="noreferrer"` implies `rel="noopener"`.

## Solution

I removed the obligation to use both `noopener` and `noreferrer` when `target="_blank"` is used, and only kept the rule to add `rel="noreferrer"`.


JoshuaKGoldberg:
Hi @esuau, thanks for sending this! Per discussion in the linked issue, it's unclear which browsers support the better behavior of inferring proper tags. We'll want to have a rule option to toggle this behavior.

I'll leave this PR open for a bit in case you have time to add that option in. If not, no worries!

esuau:
Hello. Merge was clean actually. I just did a [rebase](https://github.com/Microsoft/tslint-microsoft-contrib/pull/574/commits/7b539ba09e3f7527c4598d28c7d6f7055c712138) on your master branch. 😊

Issue #491, opened 2018-08-13
Enable more TSLint rules from core


JoshuaKGoldberg:
Keeping a pending list on this issue of rules that would be nice to enable internally for tslint-microsoft-contrib. For each rule, it'd be great to see a single PR that enables it in `tslint.json` and fixes any resultant failures in source code.

 - [x] ban-types
 - [ ] deprecation
 - [ ] jsdoc-format
 - [ ] member-access
 - [x] no-any
 - [ ] no-boolean-literal-compare
 - [ ] no-inferrable-types 
 - [ ] no-internal-module / no-namespace
 - [x] no-non-null-assertion
 - [ ] no-unbound-method
 - [ ] no-unsafe-any
 - [ ] prefer-for-of
 - [ ] prefer-function-over-method
 - [x] prefer-readonly
 - [ ] restrict-plus-operands
 - [ ] strict-boolean-expressions
 - [ ] strict-type-predicates

reduckted:
It looks like the current grunt task doesn't specify the TypeScript project, which means no type-checked rules are used. There are three warnings that pop up:

```
Running "tslint:prod" (tslint) task
Warning: The 'no-cookies' rule requires type information.
Warning: The 'no-use-before-declare' rule requires type information.
Warning: The 'prefer-readonly' rule requires type information.
```

If you run with type checking, the 'no-use-before-declare' and 'prefer-readonly' rules produce quite a lot of violations.

JoshuaKGoldberg:
Hmm, that was probably set up before TSLint even had type checking. +1.

reduckted:
It also looks like the grunt task isn't linting everything that it could. The `no-null-keyword` rule has been turned on, and running `npm test` results in everything passing, but if you run tslint directly, there are a handful of cases where the `no-null-keyword` rule finds problems. In any case, those instances of using null have been fixed indirectly by #570, and the linting via grunt will no doubt end up being fixed in #484.

reduckted:
@JoshuaKGoldberg, not sure if you want to track it here as well, but I've just disabled `no-use-before-declare` (tracked at #603) as a result of turning on type-checking in #604.

JoshuaKGoldberg:
Awesome, yes please! I'll add them to the list.

JoshuaKGoldberg:
Now that the tooling is a bit more fleshed out, marking this as a good first issue and easy difficulty. Updating the OP to be a bit more clear.

[...]

Issue #489, opened 2018-08-13
Switch from the custom unit test harness to TSLint's --test


JoshuaKGoldberg:
It'd be nice to not have to maintain the custom unit test framework, and instead use the same one as TSLint core's.

IllusionMH:
I'd like to take on this.

IllusionMH:
Update on this issue.

After #584 and #616 landed, now all infrastructure tasks are ready to accomplish test migration.

This repo has 92 rules and it will definitely take time to properly migrate all tests. I will do it in several PRs.

**If you want to help - you are welcome**. Plase leave a comment with list of rules you want to migrate, so we won't do same work.

Progress: **1 of92**

🏃  marks rules that are in batch that I'm working on.

- [ ] `chai-prefer-contains-to-index-of`
- [ ] `chai-vague-errors`
- [ ] :running_man: `export-name`
- [ ] `function-name`
- [ ] `import-name`
- [ ] `informative-docs`
- [ ] `insecure-random`
- [ ] `jquery-deferred-must-complete`
- [ ] `max-func-body-length`
- [ ] `missing-jsdoc`
- [ ] `missing-optional-annotation`
- [ ] `mocha-avoid-only`
- [ ] `mocha-no-side-effect-code`
- [ ] `mocha-unneeded-done`
- [ ] `no-backbone-get-set-outside-model`
- [ ] `no-banned-terms`
- [ ] `no-constant-condition`
- [ ] `no-control-regex`
- [ ] :running_man: `no-cookies`
- [ ] `no-delete-expression`
- [ ] `no-disable-auto-sanitization`
- [ ] `no-document-domain`
- [ ] `no-document-write`
- [ ] `no-duplicate-case`
- [ ] `no-duplicate-parameter-names`
- [ ] `no-empty-interfaces`
- [ ] `no-empty-line-after-opening-brace`
- [ ] `no-exec-script`
- [ ] `no-for-in`
- [ ] :running_man: `no-function-constructor-with-string-args`
- [ ] :running_man: `no-function-expression`
- [ ] `no-http-string`
- [ ] `no-increment-decrement`
- [ ] `no-inner-html`
- [ ] `no-invalid-regexp`
- [ ] `no-jquery-raw-elements`
- [ ] `no-missing-visibility-modifiers`
- [ ] `no-multiline-string`
- [ ] `no-multiple-var-decl`
- [ ] :running_man: `no-octal-literal`
[...]

PR #584, opened 2018-10-25
Migrate from Grunt to npm scripts

### LINKED ISSUE #484: Replace Grunt dependency with npm scripts
Grunt is a bit hefty as a dependency. It'd be nice to move away from it and have simpler `npm run *` scripts instead.

IllusionMH:
Fixes #484 

This PR reimplements current Grunts tasks with npm scripts with only minor changes to code (mostly ES2015 features and lint fixes) without any change of existing logic.

At this point new npm scripts based build should produce identical state of `dist` folder in comparison with Grunt based build (with exception of updated REAMDE.md, `package.json` that now contains additional entries in `scripts`, maps for files with lint fixes, and removed extra line endings in templates)

New benefits:

- No more problems with linting of `src/test` where `extends` config property was ignored (as mentioned by **reduckted** in #491 )
- Has starting point for #489 
- Create rule now creates tests in TSLint format
- Less dev dependencies
- no more global Grunt installation

Things that were "lost" (not re-implemented or there is no clear solution with npm scripts):

- Total build time and splits by tasks indication
- `--force` option to continue even if error happens
- ~Colorful output for warnings/errors in custom tasks (validations, generation etc.)~
- Watcher that launches tests and lint after changes (will return soon)

Additional changes are added for Debug tasks for VS Code and instructions on  how to debug it in general case.

I'd like to discuss questions about not implemented parts and changes in build steps/logic during chat.


build-tasks/common.js
```diff
@@ -0,0 +1,141 @@
+const path = require('path');
```
IllusionMH: I have no idea for better file name. Any suggestion?

build-tasks/create-rule.js
```diff
@@ -0,0 +1,78 @@
...
+
+function createTestFiles() {
+    const testContent = '// Code that should be checked by rule';
+    const tslintContent = {
+        rules: {
+            [ruleName]: true
+        }
+    };
+
+    fs.mkdirSync(testsFolder);
```
IllusionMH: It should create folder with `0o777` permission according to Node.js [docs](http://nodejs.cn/doc/node/fs.html#fs_fs_mkdir_path_mode_callback).
Is it OK or what mode value should be used?

README.md
```diff
@@ -166,28 +166,57 @@ Formatter Name          | Description | Since
...
-Then run:
+**Node.js**
 
-    node-debug grunt mochaTest
+Starting from Node.js v6.3 has built-in debugger that can be used with Chrom DevTools. For earlier versions use [`node-inspector`](https://www.npmjs.com/package/node-inspector) package.
 
-The `node-debug` command will load Node Inspector in your default browser (works in Chrome and Opera only).
+To debug rules that have Mocha tests use:
+
+    node --inspect-brk ./node_modules/mocha/bin/_mocha --no-timeouts --colors "dist/src/tests/**/*.js"
```
JoshuaKGoldberg: Nit: would you mind using <code>```</code> formatting here? GitHub's renderer thinks the `_` before `mocha` is an italic block starter.

<pre><code>```shell
node ...
```
</code></pre>

![image](https://user-images.githubusercontent.com/3335181/47511794-f4235000-d848-11e8-93a7-8fdf25c36d25.png)

File: README.md [hunk 4]
```diff
@@ -166,28 +166,57 @@ Formatter Name          | Description | Since
...
+
+Then open [chrome://inspect/](chrome://inspect/), click on `inspect` link for proper target in Remote Target section and add `tslint-microsoft-contrib` folder to Workspace (in Sources tab).
 
 Set a breakpoint somewhere in your code and resume execution. Your breakpoint should be hit.
 
+You can use `npm start` watcher that will rebuild TS files from `src` before launching debug commands.
+
+> NOTE: Run `npm test` before `npm start` to copy all required files to `dist` folder.
+
+> NOTE: If breakpoins are not hit you can try to use `inlineSourceMaps` instead of `sourceMaps` in `tsconfig.json`
```
JoshuaKGoldberg: Nit: breakpoin**t**s

build-tasks/common.js
JoshuaKGoldberg: How about splitting into `common/files.js` for file IO and `common/metadata.js` for everything else?

build-tasks/common.js
```diff
@@ -0,0 +1,141 @@
...
+const fs = require('fs');
+const glob = require('glob');
+
+function readFile(fileName) {
+    try {
+        return fs.readFileSync(String(fileName), {encoding: 'utf8'});
+    } catch (e) {
+        console.error(`Unable to read file: ${String(fileName)}. Error code: ${e.code}`);
+        process.exit(1);
```
JoshuaKGoldberg: I'm under the impression we should prefer `process.exitCode = 1;`: https://nodejs.org/api/process.html#process_process_exit_code. Thoughts?

build-tasks/common.js
```diff
@@ -0,0 +1,141 @@
+const path = require('path');
+const fs = require('fs');
+const glob = require('glob');
+
+function readFile(fileName) {
+    try {
+        return fs.readFileSync(String(fileName), {encoding: 'utf8'});
+    } catch (e) {
+        console.error(`Unable to read file: ${String(fileName)}. Error code: ${e.code}`);
```
JoshuaKGoldberg: Nit: is the `String(...)` necessary? Can we just do <code>Unable to read file: ${fileName}. ...`</code>?

tests/no-relative-imports/allow-siblings/test.js.lint
```diff
@@ -0,0 +1,7 @@
+import OfficeApp from 'OfficeApp';
```
JoshuaKGoldberg: Are these test file additions meant to be here? They seem unrelated, no?

package.json
```diff
@@ -37,23 +37,37 @@
...
+    "lint:rules": "tslint -r dist/src -t verbose -c tslint.json -e \"src/tests/**\" \"src/**/*.ts\"",
+    "lint:tests": "tslint -r dist/src -t verbose -c src/tests/tslint.json -e src/tests/references.ts \"src/tests/**/*.ts\"",
+    "validate:documentation": "node build-tasks/validate-documentation.js",
+    "validate:config": "node build-tasks/validate-config.js",
+    "generate:recommendations": "node build-tasks/generate-recommendations.js",
+    "generate:default-tslint-json": "node build-tasks/generate-default-tslint-json.js",
+    "generate:sdl-report": "node build-tasks/generate-sdl-report.js",
+    "generate:rule-metadata": "node build-tasks/generate-rule-metadata.js",
+    "generate:package-json-for-npm": "node build-tasks/generate-package-json-for-npm.js",
+    "create-rule": "node build-tasks/create-rule.js"
```
JoshuaKGoldberg: Would you mind alphabetizing these rules? I always find it really difficult to scan through lists of tasks unless they're grouped nicely.

build-tasks/create-rule.js
```diff
@@ -0,0 +1,78 @@
+const common = require('./common');
```
JoshuaKGoldberg: > const common = require('./common');

I'm not a big fan of `import * as` exports for collections of miscellaneous functions. It becomes a bit harder to tease them apart & refactor when they're prefixed with a name like that. Let's switch to `const { specificOneIWant, otherFunction ] = require('./common');`.

README.md
IllusionMH: I usually do this, but decided to follow already used style. Will change.

package.json
IllusionMH: It is possible to sort alphabetically by group/single tasks.
But current order inside "group" (e.g. `generate:<task>`) helps keep `test` script short enough.
As I've seen - `npm-run-all` preserves tasks order in case when invoked with as `npm-run-all generate:*` where order matters. Otherwise full script names should be specified in correct order.

Should I leave order inside group as per execution order (`create-rule` will be moved up, `test` & `start` down etc.) or should I reorder everything and explicitly set order inside `test` script?

tests/no-relative-imports/allow-siblings/test.js.lint
IllusionMH: `tests` folder is demo of how TSLint can be used to test rules.

This is test for `no-relative-imports` rule which supports JS files too (as [mentioned](https://github.com/Microsoft/tslint-microsoft-contrib/pull/508#issuecomment-427144264) in previous PR)
For initial example I decided to include tests for both supported languages.

Should I remove `.js.lint` files?

package.json
JoshuaKGoldberg: Leaving order inside the groups seems good, thanks!

tests/no-relative-imports/allow-siblings/test.js.lint
JoshuaKGoldberg: That seems fine. Just wanted to make sure this was intentional 😊

build-tasks/common.js
IllusionMH: I've played a bit with this and looks like `process.exitCode = 1;` would make sense only in in some checks that are performed at the end of file like in [`validate-config`](https://github.com/Microsoft/tslint-microsoft-contrib/pull/584/files#diff-959182fc06ca4da489fed83e7951391bR60) and [`validate-documentation`](https://github.com/Microsoft/tslint-microsoft-contrib/pull/584/files#diff-410f991c8117c6f5aca2760154ec2de4R33) where script execution will be finished right after this check and nothing depends on result.

In contrast `readFile` most likely be a first call in file and other code depends on its result. If there won't be implicit termination with `process.exit(1)`  - execution will continue and user will see `TypeError: Cannot read property 'toUpperCase' of undefined` or something similar.

To keep things similar across different cases I think it is better to use `process.exit(1)` in all places.

IllusionMH:
I've fixed all comments and resolved all that we decided to leave as is (exit code, `js.lint` files).

Colors are back - validation errors have yellow color, other errors that prevent further execution are red.

As for watch mode - if it is not critical for this PR, I'll make separate PR a bit later.

build-tasks/create-rule.js
JoshuaKGoldberg: Yeah this should be fine.

PR #591, opened 2018-10-26
[import-name] Fix #378

### LINKED ISSUE #378: misnamed-import warning for import A from '.'
Here's a somewhat edge-case example of what I've stumbled on:

```typescript
// file path: <proj-path>/src/Logger/utils.ts
import Logger, {
  LogItem,
} from '.'; //this refers to <proj-path>/src/Logger/index.ts
```

Here's what I get:

```
error: Logger/logger.utils.ts
severity: 'Error'
message: 'Misnamed import. Import should be named '.' but found 'Logger' (import-name)'
at: '2,1'
source: 'tslint'
```

I wrote a tslint issue on this(https://github.com/palantir/tslint/issues/2878) but was told to write the issue here.

ps: I can do a walk-around using:

```typescript
import {
  default as Logger,
  ...
} from '.';
```

but I don't like doing tricks to please Mr Linter :(

Igorbek:
Fixes #378
- ignore modules that have dotted paths like `.`, `..`, `./`

### DIFF
--- a/src/tests/ImportNameRuleTests.ts
+++ b/src/tests/ImportNameRuleTests.ts
@@ -220,4 +220,26 @@ describe('importNameRule', () : void => {
             }
         ]);
     });
+
+    it('should pass on index path modules', () => {
+        const script = `
+            import AnyName = require('.');
+            import AnyName = require('..');
+            import AnyName = require('./');
+            import AnyName = require('../');
+        `;
+
+        TestHelper.assertViolations(ruleName, script, [ ]);
+    });
+
+    it('should pass on index path ES6 modules', () => {
+        const script = `
+            import AnyName from '.';
+            import AnyName from '..';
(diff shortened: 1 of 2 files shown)


src/tests/ImportNameRuleTests.ts
```diff
@@ -220,4 +220,26 @@ describe('importNameRule', () : void => {
             }
         ]);
     });
+
+    it('should pass on index path modules', () => {
+        const script = `
+            import AnyName = require('.');
+            import AnyName = require('..');
+            import AnyName = require('./');
+            import AnyName = require('../');
```
sindresorhus: How about `import AnyName = require('../..');`?

JoshuaKGoldberg: Good point - filed #594

PR #592, opened 2018-10-26
Fix Alt-text cannot be an image file name #217

### LINKED ISSUE #217: Update react-a11y-img-has-alt - Alt-text cannot be an image file name
Alt-text cannot be an image file name (http://oaa-accessibility.org/wcag20/rule/27/, https://github.com/globant-ui/arialinter/blob/master/lib/rules/textAlternatives.js)

Dgaduin:
Fixes #217, by adding a regex to check for most common file formats in use. 
Can use some feedback on the error text and any additional formats that might have been missed

### DIFF
--- a/src/reactA11yImgHasAltRule.ts
+++ b/src/reactA11yImgHasAltRule.ts
@@ -35,6 +35,10 @@ export function getFailureStringEmptyAltAndNotEmptyTitle(tagName: string): strin
 its title attribute is not empty. Remove the title attribute.`;
 }
 
+export function getFailureStringAltIsImageFileName(tagName: string): string {
+    return `The value of alt attribute in <${tagName}> tag is an image file name. Give meaningful value to the alt attribute `;
+}
+
 /**
  * Enforces that img elements have alt text.
  */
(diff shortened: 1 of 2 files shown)


src/reactA11yImgHasAltRule.ts
```diff
@@ -114,7 +118,8 @@ class ImgHasAltWalker extends Lint.RuleWalker {
             const allowNonEmptyAltWithRolePresentation: boolean = options.length > 1
                 ? options[1].allowNonEmptyAltWithRolePresentation
                 : false;
-
+            const imageFileNameRegex: RegExp = new RegExp('^.*\\.(jpg|bmp|jpeg|jfif|gif|png|tif|tiff)$', 'i');
```
JoshuaKGoldberg: Seems odd to always recreate the regexp here each time. Could you please move this outside of the class?

Even if the underlying JS engine optimizes it away, it's a bit confusing reading through the code.

JoshuaKGoldberg:
> Can use some feedback on the error text and any additional formats that might have been missed

To be more clear: I think they're good as is.

PR #598, opened 2018-10-27
Fixes #280: Implements acessibility rule for required input elements

### LINKED ISSUE #280: new a11y rule: Required input elements must define aria-required="true"
Required input elements must define aria-required="true" 

http://www.clarissapeterson.com/2012/11/html5-accessibility/

jcnsilva:
Creates the rule react-a11y-required to assert that required input elements have the aria-required attribute set to true.

Fixes #280.


Harley-Adams:
Looks good to me. I think with modern browsers it's redundant but I don't have a source on that and it's been awhile since I've been in the area.

PR #600, opened 2018-10-29
Remove invalid download URL

mesaugat:
npm-5.2.2 tree doesn't exist

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,8 @@ A set of [TSLint](https://github.com/palantir/tslint) rules used on some Microso
 npm install tslint-microsoft-contrib --save-dev
 ```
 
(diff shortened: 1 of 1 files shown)


mesaugat:
Looks like there is a documentation check in place for the current version. The latest version available in npm is `5.2.1` but the README.md has `npm-5.2.2` and so does `package.json`.

A release might be required.

Please close this ticket if the download URL is required in the README.md. Sorry for the trouble!

If the download tree is not required, I can remove the version check from `validate-documentation.js`

JoshuaKGoldberg:
Thanks @mesaugat! Looks like this was an oversight on my part; I should have uploaded that branch. Will do.

PR #609, opened 2018-10-29
Increased test timeout to 5000s

### LINKED ISSUE #606: Test timeout should be increased beyond 2s
As seen in #601, the Windows builds sometimes time out... ☹

JoshuaKGoldberg:
Fixes #606


IllusionMH:
Nice! 👍

PR #612, opened 2018-10-29
Extended tslint.json from tslint:all

### LINKED ISSUE #608: Extend tslint.json from tslint:all
This will require disabling a bunch of rules, many of which will be listed in #491. Our `tslint.json` predates the ability to extend configurations.

JoshuaKGoldberg:
Changes `getRuleNames` utility to `getContribuRuleNames` and no longer retrieve core TSLint rules. Also adds comments and sections in our `tslint.json` to help distinguish classes of rule settings.

Fixes #608.


build-tasks/validate-config.js
```diff
@@ -2,62 +2,35 @@
...
 
 const errors = [];
-getAllRuleNames().forEach(ruleName => {
-    if (rulesToSkip.has(ruleName)) {
+getContribRuleNames().forEach(ruleName => {
+    if (disabledRules.has(ruleName)) {
         return;
     }
+
     if (tslintConfig.rules[ruleName] !== true && tslintConfig.rules[ruleName] !== false) {
```
IllusionMH: There is still a problem with this logic.
Looks like it won't allow to do `"rule": [false, "param"]` but still pass on `"rule": false`.
Maybe it makes sense to update logic with whole rule?

JoshuaKGoldberg: Agreed, this is a good time.

IllusionMH:
👍

build-tasks/validate-config.js
JoshuaKGoldberg: @IllusionMH does this look good to you?

IllusionMH: Yes, this should cover all necessary cases for this repo (no rules with different severity).

PR #616, opened 2018-10-30
Add watcher that will run tests and lint

IllusionMH:
This PR is continues efforts to replace Grunt with npm scripts #484 and adds watcher that wasn't implemented in scope of #584.

New script `watch:run-test` added that will watch compiled `.js` files in `dist`, `test-data` and `test` folders content and run tests and lint tasks for repo.

Custom implementation used instead of `chokidar-cli` because later doesn't wait before command will be finished before starting new one, which might lead to running 2-3 parallel Mocha tests.

Possible future improvements:
* Add `--skip-initial` flag that will skip initial run of tests (when watcher ready)
* Add `--restart-on-change` flag that will stop currently running tests if watched file is changes (ATM blocked at https://github.com/mysticatea/npm-run-all/issues/51)

Also contains fixes for inconsistent quotes in other JS files. Remove from this PR?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1237,6 +1237,18 @@ More information about writing rule tests can be found in [TSLint documentation]
 
 Please also see our [Code of Conduct](CODE_OF_CONDUCT.md).
 
+### Watchers
+
+Two npm scripts can be used to watch files and simplify testing and linting with constant changes.
+
+Use `npm start` to start watcher that will rebuild rules and Mocha tests.
+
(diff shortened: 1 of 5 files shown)


JoshuaKGoldberg:
> Remove from this PR?

Normally I'd say yes, but since they're in the same area and the changes aren't functional, seems fine for now.

Would you mind filing issues for followups you'd like to see? I have a few ideas but don't want to trample over yours first 🙂.

PR #629, opened 2018-11-03
Disallow current/parent folders inside module path

### LINKED ISSUE #624: Enhancement: Make `no-relative-imports` rule stricter
At this moment `no-relative-imports` dissalows `..` only at the beginning of the import path (see #508).

Since next release most likely will be [major release](https://github.com/Microsoft/tslint-microsoft-contrib/issues/585) it is good opportunity to make "breaking" change disallow `..` inside path.

This will be easy check and won't require a lot of effort.

Ideally - disallow not only `/../` but also `/./` inside of path.

Update rule instead of new one - because it will be basically copy-paste with minor changes in check function (that shouldn't conflict with existing) and error message.

IllusionMH:
Fixes #624 

This PR disallows references to current and parent directory (e.g. `common/../other-folder/App`).
This check doesn't depend on `allow-siblings` argument, and new error will be shown only if path passes checks for relative or parent references at the beginning.

I've removed Mocha tests for `no-relative-imports` rule, so this is first rule that fully migrated to TSLint format tests.


JoshuaKGoldberg:
> first rule that fully migrated to TSLint format tests.

Whoohoo!

IllusionMH:
Thanks!
To be honest - I was too lazy to add all cases to mocha tests, and just updated existing TSLint bases tests 😭

JoshuaKGoldberg:
Ha, more evidence that the bases tests are the way to go for easier contributing! 😉

PR #630, opened 2018-11-04
Rename license.txt to LICENSE and update it's contents

mesaugat:


JoshuaKGoldberg:
Thanks for the change, @mesaugat, but is there any reason behind the change? It's better to not make changes to these files without a good reason.

mesaugat:
@JoshuaKGoldberg There's no such reason but if you view the file, GitHub will show a [preview of the license on the header](https://github.com/Microsoft/vscode/blob/master/LICENSE.txt). Also, most repositories use the uppercase `LICENSE` term. Just wanted to be consistent.

Issue #577, opened 2018-10-24
Remove our custom fix formatters


JoshuaKGoldberg:
`fix-no-require-imports` and `fix-no-var-keyword` predate TSLint's ability to `--fix` violations. They're no longer necessary and should be marked as deprecated. 

After the next major release following their deprecation, they should be removed.

JoshuaKGoldberg:
☠️  *It's time!* ☠️ 

Per #876, this repository is no longer accepting feature pull requests. TSLint is being deprecated and we recommend you switch to https://typescript-eslint.io.

Thanks for open sourcing with us, everyone!

PR #638, opened 2018-11-05
Removed mentions of now-deprecated formatters from README.md

JoshuaKGoldberg:
Continues #577.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1197,15 +1197,6 @@ There certainly are a lot of options! Here are some links to get you started.
   </tbody>
 </table>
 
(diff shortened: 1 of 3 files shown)


IllusionMH:
Just to clarify: Since they are deprecated, and 6.0 will be first release with deprecation warning, shouldn't in stay in docs until removal with `deprecated` label? Or you want to remove that from docs to avoid unintentional use (because some will miss `deprecated label)?

JoshuaKGoldberg:
> remove that from docs to avoid unintentional use

Yes, that.

PR #641, opened 2018-11-05
Remove absolute path from rules error messages.

### LINKED ISSUE #634: Proposal: Remove absolute paths from error messages (ideally before 6.0)
Absolute paths block micration to TSLint test format, since error message will depend on file location, and can't be universal across Windows/Linux/Mac OS.
File name/path should be in formatter output (I've checked all core formatters and they provide absolute path).

There are 3 rules that use absolute path in its erros:

- `export-name` [source](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/exportNameRule.ts#L168) - use file base name - will be easy to  compare.
- `missing-jsdoc` [source](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/missingJsdocRule.ts#L40) (already deprecated in this repo) - remove absolute path and change `:` to `.` at the end of error message.
- `react-no-dangerous-html` - [source](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/reactNoDangerousHtmlRule.ts#L101) - remove absolute path and `of source file` string before it.

What do you think? How critical to have absolute path here (especially last one)?

As for me - it will make error messages hard to read, because path can be too long.
In IDE I'm already in that file and in CI formatter will output path relative to

IllusionMH:
#### PR checklist

-   [X] Addresses an existing issue: fixes #634 
-   [X] ~New feature,~ bugfix, ~or enhancement~
    -   [X] Includes tests
-   [ ] Documentation update

#### Overview of change:

- `export-name` now has only base file name in error message.
- `missing-jsdoc` removed filename from error message
- `react-no-dangerous-html` filename from error message

I haven't changed how string concatenations is made for multiline errors because anyway they will be migrated soon.

Added inline comment.

This PR with template looks better and important :)

### DIFF
--- a/src/tests/ReactNoDangerousHtmlRuleTests.ts
+++ b/src/tests/ReactNoDangerousHtmlRuleTests.ts
@@ -22,9 +22,8 @@ function someFunction() {
             [
                 {
                     failure:
-                        'Invalid call to dangerouslySetInnerHTML in method "<unknown>"\n    of source file ' +
-                        Utils.absolutePath('file.tsx') +
-                        '"\n    Do *NOT* add a suppression for this warning. ' +
+                        'Invalid call to dangerouslySetInnerHTML in method "<unknown>".' +
+                        '\n    Do *NOT* add a suppression for this warning. ' +
                         'If you absolutely must use this API then you need\n    to review the usage with a security expert/QE ' +
                         'representative. If they decide that this is an\n    acceptable usage then add the exception ' +
                         'to xss_exceptions.json',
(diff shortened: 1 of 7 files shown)


src/tests/ReactNoDangerousHtmlRuleTests.ts
```diff
@@ -34,9 +33,8 @@ function someFunction() {
                 },
                 {
                     failure:
-                        'Invalid call to dangerouslySetInnerHTML in method "<unknown>"\n    of source file ' +
-                        Utils.absolutePath('file.tsx') +
-                        '"\n    Do *NOT* add a suppression for this warning. ' +
+                        'Invalid call to dangerouslySetInnerHTML in method "<unknown>".' +
+                        '\n    Do *NOT* add a suppression for this warning. ' +
```
IllusionMH: Added `.` to the end of previous line. And removed non-matched `"` after file name (probably copy-paste os string with `\n`)

PR #642, opened 2018-11-05
Make export-name RegExp more strict

### LINKED ISSUE #632: Incorrect RegExp in export-name rule leads to superfluously permissive
As was discovered in https://github.com/Microsoft/tslint-microsoft-contrib/pull/623#discussion_r230436742 RegExp in `export-name` rule has issue with improperly escaped `.` as well as missing check for line start/end. These problems leads to several problems.

For class name `SomeClass` it allows:
1. Wrong `.` escape: `SomeClassButAnyTextHere.tsx`, `SomeClassWithoutExtension`
1. No end of line check: `SomeClass.ts.or.not`
1. No start of line check: `LooksLikeSomeClass.ts`
1. Everything combined `NotSureThatSomeClassIsSomewhereInThis.long.filename`

According to description I would expect only `SomeClass.ts` or at least `SomeClass.suffix.tsx` (but better to make it as option).

What do you think about cases above and which should be disallowed by this rule?

IllusionMH:
#### PR checklist

-   [X] Addresses an existing issue: fixes #632 
-   [X] New ~feature,~ bugfix, ~or enhancement~
    -   [X] Includes tests
-   [ ] Documentation update

#### Overview of change:

This PR requires exported class name strictly match start of file name followed by `.` and atleast 1 more extension character.

Will rebase/cleanup after #641 merged

Ignore `Rule` class name from checks in this repo, because all rule files exports just `Rule` which is violation.


src/tests/ExportNameRuleTests.ts
```diff
@@ -257,14 +257,28 @@ describe('exportNameRule', (): void => {
                 {
                     failure:
                         'The exported module or identifier name must match the file name. ' +
-                        'Found: test-data/ExportName/ExportNameRuleFailingTestInput2.tsx and ThisIsNotTheNameOfTheFile',
+                        'Found: test-data/ExportName/ExportNameRuleFailingTestInput2.tsx and ExportNameRuleFailingTestInput',
```
IllusionMH: Now checks that name of exported class is immediately followed by `.`

tslint.json
```diff
@@ -60,7 +60,7 @@
         "chai-prefer-contains-to-index-of": true,
         "chai-vague-errors": true,
         "encoding": true,
-        "export-name": true,
+        "export-name": [true, { "allow": ["^Rule$"] }],
```
IllusionMH: Quick fix instead of disabling/enabling check in 80 files, which is better for separate PR.

PR #647, opened 2018-11-06
Prepared 6.0.0-beta release

JoshuaKGoldberg:
🚀!

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,14 +15,14 @@ npm install tslint-microsoft-contrib --save-dev
 
 ## TSLint and corresponding tslint-microsoft-contrib version
 
-| TSLint version | tslint-microsoft-contrib version              |
-| -------------- | --------------------------------------------- |
-| **>= 5.x**     | 5.x (supporting TypeScript 2.3.x, >=2.4, 3.x) |
-| **>= 4.x**     | 4.x (supporting TypeScript 2.1.x)             |
-| **>= 3.2.x**   | 2.x                                           |
-| **3.1.x**      | unsupported                                   |
-| **3.0.x**      | unsupported                                   |
-| **2.x**        | 1.x                                           |
+| TSLint version | tslint-microsoft-contrib version |
+| -------------- | ---------------------------------------------**\_\_\_\_** |
+| **>= 5.x** | 5.x and 6.x (supporting TypeScript 2.3.x, >=2.4, 3.x) |
+| **>= 4.x** | 4.x (supporting TypeScript 2.1.x) |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -15,14 +15,14 @@ npm install tslint-microsoft-contrib --save-dev
...
-| TSLint version | tslint-microsoft-contrib version              |
-| -------------- | --------------------------------------------- |
-| **>= 5.x**     | 5.x (supporting TypeScript 2.3.x, >=2.4, 3.x) |
-| **>= 4.x**     | 4.x (supporting TypeScript 2.1.x)             |
-| **>= 3.2.x**   | 2.x                                           |
-| **3.1.x**      | unsupported                                   |
-| **3.0.x**      | unsupported                                   |
-| **2.x**        | 1.x                                           |
+| TSLint version | tslint-microsoft-contrib version |
+| -------------- | ---------------------------------------------**\_\_\_\_** |
```
mesaugat: `**\_\_\_\_**` ??

JoshuaKGoldberg: Great question...

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tslint-microsoft-contrib",
-  "version": "5.2.1",
+  "version": "6.0.0-beta1",
```
IllusionMH: All mentions above and in title are `6.0.0-beta0`

JoshuaKGoldberg: Thinking about this a bit - I doubt there'll be multiple betas, so I'll just switch it to `6.0.0-beta`.

package-lock.json
```diff
@@ -3371,7 +3371,7 @@
     },
     "readable-stream": {
       "version": "2.3.6",
-      "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
+      "resolved": "http://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
```
IllusionMH: This change looks a bit strange. Otherwise looks good.

IllusionMH:
Probably there should be some guide in contributing docs, how to add next version in readme. For example `@next` (`v.next` or `%next%`) which can be just search/replace action before release.

JoshuaKGoldberg:
@IllusionMH agreed, that's a good idea.

package-lock.json
JoshuaKGoldberg: Yeah I don't know why it was happening 😕 but it happened across multiple machines, so I just let it in.

PR #650, opened 2018-11-08
Upgrade prettier version to 1.15.0

mesaugat:
#### PR checklist

- [x] Enhancement

#### Overview of change:

https://prettier.io/blog/2018/11/07/1.15.0.html

Prettier just came out with a new version. Because we upgraded to prettier a few days back, I think we are in the position to upgrade to the latest prettier version as of now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3298,9 +3298,9 @@
       "dev": true
     },
     "prettier": {
-      "version": "1.14.3",
-      "resolved": "https://registry.npmjs.org/prettier/-/prettier-1.14.3.tgz",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -3371,7 +3371,7 @@
     },
     "readable-stream": {
       "version": "2.3.6",
-      "resolved": "http://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
+      "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.3.6.tgz",
```
mesaugat: ¯\\_(ツ)_/¯

JoshuaKGoldberg: Haha this refused to be `https` on my machine... fine!

PR #651, opened 2018-11-10
#274 Implement react-a11y-accessible-headings

### LINKED ISSUE #274: new a11y rule: accessible headings
All heading elements (h1..h6) should have text content
    • https://oaa-accessibility.org/rule/39/
    • https://github.com/globant-ui/arialinter/blob/master/lib/rules/headingsAndLabels.js

Text content for a heading must not come just from image alt text. (http://oaa-accessibility.org/wcag20/rule/19/)

Heading content should be concise. (http://oaa-accessibility.org/wcag20/rule/20/)

Heading elements (h1..h6) should be used for structuring information on the page. (http://oaa-accessibility.org/wcag20/rule/23/)

Missing or empty H1 element. (http://oaa-accessibility.org/wcag20/rule/14/)

H1 element content should not come only from the alt text of an image. (http://oaa-accessibility.org/wcag20/rule/15/)

H1 should match a subset of the words in the title element. (http://oaa-accessibility.org/wcag20/rule/16/)

No more than two h1 elements. (http://oaa-accessibility.org/wcag20/rule/17/)

noamyogev84:
#### PR checklist
- Fixes #274 
- New feature
- Includes tests
- Documentation update

#### Overview of change:
 **The following requirements are met**:
 1. All heading elements (h1..h6) should have text content.
 2. Heading content should be concise (configurable rule)
 3. Heading elements (h1..h6) should be used for structuring information on the page. (implemented according to #274 discussion)
 4. H1 should match a subset of the words in the title element.
 5. No more than two h1 elements.


noamyogev84:
Hi @JoshuaKGoldberg. Care to look at my PR?

Thanks!!

src/reactA11yAccessibleHeadingsRule.ts
```diff
@@ -0,0 +1,164 @@
...
+import * as Lint from 'tslint';
+
+import {ExtendedMetadata} from './utils/ExtendedMetadata';
+
+const BAD_ORDER_HEADING_FAILURE_STRING: string = 'Heading elements should be used for structuring information on the page';
+const EMPTY_HEADING_FAILURE_STRING: string = 'Heading elements must not be empty';
+const BAD_HEADING_LENGTH_STRING: string = 'Heading content should be concise';
+const H1_HEADING_NOT_SUBSET_OF_TITLE_STRING: string = 'H1 heading should match a subset of the words in the title element';
+const BAD_NUMBER_H1_HEADING_FAILURE_STRING: string = 'H1 heading cannot exceed 2 elements';
+const VALID_HEADING_TYPES: string[] = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
```
JoshuaKGoldberg: > `VALID_HEADING_TYPES: string[]`

Turn this into a `Set` so you can use `.has(` later on instead of a more complex `.find`.

src/tests/reactA11yAccessibleHeadingsRuleTests.ts
```diff
@@ -0,0 +1,211 @@
...
+                "name": Utils.absolutePath("file.tsx"),
+                "ruleName": "react-a11y-accessible-headings",
+                "startPosition": { "character": 30, "line": 4 }
+            }
+        ]);
+    });
+    it('should pass when H1 heading text is a substring of title text', () : void => {
+        const script : string = `
+            import React = require('react');
+            const title = <title>"I'm so great look at me..."</title>;
```
JoshuaKGoldberg: > `<title>...</title>`
.
It looks like you're keeping track of the last known title text with `captureTitleText` / `capturedTitleText`, then comparing later nodes against it. That's making two assumptions in this rule that doesn't always hold true:

* The file that contains headings will also contain the `<title>`
* Files contain nodes in order

You can't actually set the `<title>` node easily in JS apps since it's part of the `<head`>. See https://github.com/gaearon/react-document-title

I don't think there's a way around these limitations. You'll likely have to get rid of this logic ☹.

JoshuaKGoldberg: Similarly, the rule's tests should make sure that it doesn't flag files where multiple components are able to create an `<h1>` element, as it might be that only one is ever called. This should not cause a complaint:

```typescript
export const createTitleA = () => <h1>A</h1>;
export const createTitleB = () => <h1>B</h1>;
```

noamyogev84: Thanks for this review @JoshuaKGoldberg !

noamyogev84: @JoshuaKGoldberg 

I think I completely missed a bunch of stuff after reading what you're describing in your last comment. This is relevant to issues 3 and 5.
I think I need to put those tests into context and re-implement a bunch of stuff. (like distinction between assignments and actual script...)

noamyogev84: Hi @JoshuaKGoldberg.
fixed all the issues, and re implemented some rule's logic. (I hope not too naively )

src/reactA11yAccessibleHeadingsRule.ts
```diff
@@ -0,0 +1,194 @@
...
+        if (sourceFile.languageVariant === ts.LanguageVariant.JSX) {
+            return this.applyWithWalker(new ReactA11yAccessibleHeadingsWalker(sourceFile, this.getOptions()));
+        } else {
+            return [];
+        }
+    }
+}
+
+class ReactA11yAccessibleHeadingsWalker extends Lint.RuleWalker {
+    //private headings: { [id: string] : IHeadingResult; } = {};
```
JoshuaKGoldberg: Please remove commented out code. Here and one or two other places.

src/tests/reactA11yAccessibleHeadingsRuleTests.ts
```diff
@@ -0,0 +1,281 @@
+import { Utils } from '../utils/Utils';
+import { TestHelper } from './TestHelper';
+
+describe('reactA11yAccessibleHeadingsRule', (): void => {
+    const ruleName: string = 'react-a11y-accessible-headings';
+    describe('react accessible heading rules for plain heading elements', (): void => {
+        it('should pass on when heading is non empty expression', (): void => {
+            const script: string = `
```
JoshuaKGoldberg: It looks like the rule's logic is still running under the assumption that a single file contains all the `<h*>` s a particular page might have... but that's often not so. Try adding this test script:

```tsx
export const createBasicH1 = () => <h1>Hello, world!</h1>;
export const createBasicH2 = () => <h2>Hello!</h2>;
export const createTextH1 = (text: string) => <h1>{text}</h1>;
export const createTextPropsH1 = (text: { text: string }) => <h1>{text}</h1>;
export const createTextPropsH2 = (text: { text: string }) => <h2>{text}</h2>;
export const andAnotherH2 = () => <h2>What a rude edge case</h2>;
export const andAnotherH1 = () => <h1>Eeeeeeevil!</h1>;
```

Theoretical context to that snippet: other files would import those components and use them elsewhere. That snippet and the equivalent with classes extending from `React.Component` _(or something like them)_ should be included in your test suite and shouldn't flag any failures for having multiple headings of the same type.

The red flag in your implementation is that you're storing private member variables in the walker to keep track of heading information.. but you can really only keep track of those contents _within a particular component or `render()` function_.

Recommendation: make the rule much dumber and have it only consider counts, etc. within individual components. This should fail:

```tsx
export const someComponent = () => (
    <div>
        <h1>Hello,</h1>
        <h1>world!</h1>
    </div>
);
```

...but this shouldn't:

```tsx
export const createFirstH1 = () => <h1>Hello,</h1>;
export const createSecondH1 = () => <h1>world!</h1>;
```

Does that seem reasonable?

src/reactA11yAccessibleHeadingsRule.ts
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    protected visitVariableDeclaration(node: ts.VariableDeclaration): void {
+        const options: IValidationOptions = {
+            validateText: true,
+            validateOrder: false,
+            validateH1Count: false
+        };
+
+        let element = this.tryGetNode(node, ts.SyntaxKind.JsxElement);
```
JoshuaKGoldberg: Huh, interesting. Does the rule not normally visit `JsxElement`s? I'm curious if & why you needed to add this?

src/reactA11yAccessibleHeadingsRule.ts
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+
+    private validateHeadingText(headingNode: ts.JsxElement): void {
+        let headingText: string | undefined;
+        const innerNode = headingNode.children[0];
+        if (headingNode.children.length === 0) {
+            this.addFailureAt(headingNode.getStart(), headingNode.getWidth(), EMPTY_HEADING_FAILURE_STRING);
+        } else {
+            if (innerNode.kind === ts.SyntaxKind.JsxExpression) {
```
JoshuaKGoldberg: Fun fact: for each `ts.SyntaxKind`, there should be a helper available from the `tsutils` package of the form `isSomeKind(node: ts.Node): node is SomeKind`. If you use those, you won't have to manually cast the node later on (e.g. `expressionNode` below); TypeScript will "narrow" the type of the node. Pretty nifty 😉 and lets you get away from extra variables and `.kin` checking logic such as what you're doing in `expressionNode`.

If there's any syntax kind that doesn't have a corresponding check in `tsutils`, add it in the `utils` file in this repo and file an issue on `tsutils` to add it.

src/tests/reactA11yAccessibleHeadingsRuleTests.ts
noamyogev84: @JoshuaKGoldberg sounds very reasonable and educating. 
Just as a quick update - The test snippet which you've suggested actually passes.
 
So, do you think that maybe the best approach is to remove all member variables, keep using `VisitVariableDeclaration` as the entry point, and keep implementation inside?

src/reactA11yAccessibleHeadingsRule.ts
noamyogev84: Using `VisitVariableDeclaration` and then traversing `Jsx` child nodes using `this.tryGetNode`. How would you capture the context otherwise?

src/reactA11yAccessibleHeadingsRule.ts
noamyogev84: Awesome.

noamyogev84:
HI @JoshuaKGoldberg. Care to take a look on this one?

noamyogev84:
Hey @JoshuaKGoldberg. pinging you one mroe time. :)

src/tests/reactA11yAccessibleHeadingsRuleTests.ts
```diff
@@ -0,0 +1,201 @@
+import { Utils } from '../utils/Utils';
+import { TestHelper } from './TestHelper';
+
+describe('reactA11yAccessibleHeadingsRule', (): void => {
```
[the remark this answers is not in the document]
noamyogev84: Hi @JoshuaKGoldberg. Would you care to elaborate on the above example? 
When I explore your example in `ts-ast-viewer` i see that i can "scope" these jsx expressions in several ways (e.g `VisitClasssDeclaration`, `VisitMethodDeclaration`...). let's say I choose to use `VisitMethodDeclartaion` - is this sufficient? what if within the method I'll have different variable declarations, hence i'm in the first case of my implementation again... 

What do you think is the best approach here?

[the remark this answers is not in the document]
JoshuaKGoldberg: Oh I just meant that your tests file doesn't test any components that extend `React.Component`. You've got a bunch of tests for functional components but no classes.

noamyogev84:
Hi @JoshuaKGoldberg. Hope all the issues in your review are solved now. I hope there's enough weird cases covered in my unit tests. please feel free to review again.. :)

src/tests/reactA11yAccessibleHeadingsRuleTests.ts
```diff
@@ -0,0 +1,365 @@
...
+            const someComponent3  = () => <h1>test</h1>;
+
+        `;
+
+        TestHelper.assertViolations(ruleName, script, []);
+    });
+    it('should fail when heading elements numbering increase by more then one level consecutively - react component', (): void => {
+        const script: string = `
+            import React = require('react');
+            export class DontWeJustLoveMultipleApis extends React.Component {
```
JoshuaKGoldberg: > `DontWeJustLoveMultipleApis`

😂

noamyogev84:
@JoshuaKGoldberg Thanks for your patience, it was my pleasure!

PR #668, opened 2018-12-08
react-a11y-no-onchange rationale

karfau:
#### PR checklist

-   [ ] Addresses an existing issue: fixes #0000
    -   [ ] Includes tests
-   [x] Documentation update

#### Overview of change:

As part of updating https://github.com/bettermarks/bm-tslint-rules to v6 of this rule set I was checking the rules that have been added.
Because there was no reference mentioned for `react-a11y-no-onchange` I did some searching and found a page that seemed to be a good fit as a reference.

I checked and found that most other `react-a11y-*` rules have this information as part of the "supported rules" table in the README.
Because I'm the developer of https://github.com/karfau/tslint-report (that adds rules metadata to it's reports) and I like generated documentation since it reduces the burden of keeping things in sync manually, I also added the references for all `react-a11y-*` rules to the `rationale`field of their metadata.

When I found `metadata.description` didn't match the README and it was easy to do I updated the metadata accordingly. Since some of them contained markdown instead of HTML I sticked to that in the decsription field.

Along the way I also found some minor issues in the README that I fixed:
- an unclosed `li` tag
- a rule that was not at the correct place regarding to alphabetical order

* I used template strings for the rationale, to be able to just copy paste the multi-line strings. I hope it is supported in all target environments or will be transpiled accordingly?
* Currently the `rationale` contains HTML but the descriptions have markdown. I would be willing to also convert the `rationale` values to markdown if requested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -941,16 +941,6 @@ We recommend you specify exact versions of lint libraries, including `tslint-mic
     </tr>
     <tr>
       <td>
-        <code>use-simple-attributes</code>
-      </td>
-      <td>
-        Simpler attributes in JSX and TSX files helps keep code clean and readable.
-        Separate complex expressions into their own line and use clear variable names to make your code more understandable.
-      </td>
-      <td>6.0.0</td>
-    </tr>
-    <tr>
-      <td>
         <code>react-a11y-props</code>
       </td>
       <td>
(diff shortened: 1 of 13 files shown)


JoshuaKGoldberg:
Awesome, thanks for sending this in @karfau! I'm not able to merge this in right now, but once I can we'll get this merged in.

https://github.com/karfau/tslint-report looks interesting - might be useful for https://github.com/palantir/tslint/issues/2267 (TSLint > Feature request: summary report).

Issue #611, opened 2018-10-29
Remove SDL hooks metadata and build steps


JoshuaKGoldberg:
SDL refers to the [Microsoft Security Development Lifecycle](https://www.microsoft.com/en-us/sdl). It's a bit of extra work to add our own metadata on top of TSLint's. Does anybody use this?

/cc @HamletDRC

HamletDRC:
@JoshuaKGoldberg honestly no, we can stop updating this. MS kept a list of warnings in Warnings Central that we updated. But we can kill this.

JoshuaKGoldberg:
☠️  *It's time!* ☠️ 

Per #876, this repository is no longer accepting feature pull requests. TSLint is being deprecated and we recommend you switch to https://typescript-eslint.io.

Thanks for open sourcing with us, everyone!

Issue #646, opened 2018-11-06
Many rules are unnecessarily marked as typescriptOnly


JoshuaKGoldberg:
Each rule with `typescriptOnly: true` should be checked to see if it can be turned on for `.jsx?` files.

~~Similarly, `.endsWith("tsx")` from #525 should be replaced with a unified utility to check if a file name is jsx/tsx.~~ Resolved by #527.

JoshuaKGoldberg:
☠️  *It's time!* ☠️ 

Per #876, this repository is no longer accepting feature pull requests. TSLint is being deprecated and we recommend you switch to https://typescript-eslint.io.

Thanks for open sourcing with us, everyone!

PR #679, opened 2018-12-15
Used inquirer to prompt for properties when creating a new rule.

### LINKED ISSUE #258: Create a rule generator by using a micro-generator f/w
Using a micro-generator framework such as [plop](https://github.com/amwmedia/plop) would make it easy for all contributors of this repository to create new rules with a level of uniformity, and also faster, by asking a set a questions and boilerplating the appropriate code.

Also because [context switching is expensive](http://www.petrikainulainen.net/software-development/processes/the-cost-of-context-switching/) :) and [saving time is not the only benefit to automating workflows](https://medium.com/@kentcdodds/an-argument-for-automation-fce8394c14e2)

What are your thoughts about this ? Also considering time versus ROI (https://xkcd.com/1205/) ?

reduckted:
#### PR checklist

-   [x] Addresses an existing issue: fixes #258 
    -   ~Includes tests~
-   [x] Documentation update

#### Overview of change:

As per #258, I've used [Inquirer.js](https://github.com/SBoudrias/Inquirer.js) to prompt for the rule's metadata. 

All metadata except for the `options` (and related properties) and `commonWeaknessEnumeration` can be specified. I figured it wasn't worth doing anything with CWE because of #611.

The existing generator didn't actually work, because it tried to add the new rule to the root `tslint.json` file, but that file contains comments and cannot be parsed as JSON. For now, I've removed that step, but I kinda think it's something that should still be done. I'm just not sure how to go about doing it. 😕 Any suggestions are welcome.

I've also tried to be nice and open the new rule's files after they've been created. I can detect that the generator is running in the VS Code terminal by looking at the environment variables and just launch `code new_file_name.ts` to open the file in the current VS Code instance. It works for me on Windows. In theory it should work on other operating systems, but I haven't confirmed that. I'm happy to remove this feature if you don't think it's worth the trouble.

### DIFF
--- a/build-tasks/create-rule.js
+++ b/build-tasks/create-rule.js
@@ -85,17 +85,8 @@ inquirer.prompt(questions).then(answers => {
     console.log(`Source file: ${sourceFileName}`);
     console.log(`Test files:   ${testFileNames.join(', ')}`);
 
-    // If we're running in the VS Code terminal, try to open the
-    // new files. If we can't do it, then it's not a big deal.
-    if (process.env.VSCODE_CWD) {
-        try {
-            testFileNames.forEach(fileName => execSync(`code "${fileName}"`));
-            execSync(`code "${sourceFileName}"`);
-        } catch (ex) {
-            // Couldn't open VS Code.
-            console.log(ex);
-        }
-    }
+    // Attempt to open the files in the current editor.
+    tryOpenFiles([...testFileNames, sourceFileName]);
 });
 
 function createImplementationFile(answers) {
@@ -158,4 +149,32 @@ function camelCase(input) {
 
 function pascalCase(input) {
     return input.charAt(0).toUpperCase() + input.substr(1);
+}
+
+function tryOpenFiles(files) {
+    // Check if we're running in the VS Code terminal. If we
+    // are, then we can try to open the new files in VS Code.
+    // If we can't open the files, then it's not a big deal.
+    if (process.env.VSCODE_CWD) {
+        let exe;
+
+        // We need to check if we're running in normal VS Code or the Insiders version.
+        // The `TERM_PROGRAM_VERSION` environment variable will contain the version number
+        // of VS Code. For VS Code Insiders, that version will have the suffix "-insider".
+        const version = process.env.TERM_PROGRAM_VERSION || '';
+        if (version.endsWith('-insider')) {
+            exe = 'code-insiders';
+            console.log('Opening the new files in VS Code Insiders...');
+        } else {
+            exe = 'code';
+            console.log('Opening the new files in VS Code...');
+        }
+
+        try {
+            files.forEach(fileName => execSync(`${exe} "${fileName}"`));
+        } catch (ex) {
+            // Couldn't open VS Code.
+            console.log(ex);
+        }
+    }
 }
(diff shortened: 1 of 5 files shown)


build-tasks/templates/rule.template.js
```diff
@@ -1,42 +1,37 @@
...
         ruleName: '${ruleName}',
-        type: 'maintainability',    // one of: 'functionality' | 'maintainability' | 'style' | 'typescript'
-        description: '... add a meaningful one line description',
+        type: '${type}',
+        description: '${description}',
         options: null, // tslint:disable-line:no-null-keyword
         optionsDescription: '',
-        optionExamples: [],         // Remove this property if the rule has no options
+        optionExamples: [], // TODO: Remove this property if the rule has no options
         typescriptOnly: false,
```
IllusionMH: Would be nice to ask this during setup.
There is issue #646 to convert old rules. And I think that new rules should support JS from the begging if applicable.

build-tasks/create-rule.js
```diff
@@ -1,79 +1,145 @@
...
+    const sourceFileName = createImplementationFile(answers);
+    const testFileName = createTestFiles(answers.name);
+
+    console.log(`Rule '${answers.name}' created.`);
+    console.log(`Source file: ${sourceFileName}`);
+    console.log(`Test file:   ${testFileName}`);
+
+    // If we're running in the VS Code terminal, try to open the
+    // new files. If we can't do it, then it's not a big deal.
+    if (process.env.VSCODE_CWD) {
```
IllusionMH: 👍

build-tasks/create-rule.js
```diff
@@ -1,79 +1,145 @@
...
+        try {
+            execSync(`code "${testFileName}"`);
```
IllusionMH: Is there an easy way to tell if it is `code` or `code-insiders`? In my case it will open separate instance of VS Code instead of reusing current VS Code insiders window.

May be use `inquirer` to prompt if user want to open files in case when `VSCODE_CWD` present (conditionally add question to initial list)?

build-tasks/templates/rule.template.js
```diff
@@ -1,42 +1,37 @@
...
 export class Rule extends Lint.Rules.AbstractRule {
-
     public static metadata: ExtendedMetadata = {
         ruleName: '${ruleName}',
-        type: 'maintainability',    // one of: 'functionality' | 'maintainability' | 'style' | 'typescript'
-        description: '... add a meaningful one line description',
+        type: '${type}',
+        description: '${description}',
         options: null, // tslint:disable-line:no-null-keyword
         optionsDescription: '',
```
IllusionMH: While here: may be add `TODO` here too?

build-tasks/create-rule.js
reduckted: Oh, nice catch. I hadn't even thought about that. Looks like there's a `TERM_PROGRAM_VERSION` environment variable we can use to check if the version of VS Code is the Insiders version.

reduckted:
Thanks for the suggestions. I've made those extra changes. 

Detecting VS Code Insiders ended up being easier than I thought it would be. I've tested the generator in VS Code and Insiders, and the new files opened up in the correct instance, but it would be great if someone else could confirm it works for them just in case there's something special about my configuration that made it work.

JoshuaKGoldberg:
Thanks @reduckted! This is very snazzy.

PR #682, opened 2018-12-18
Convert no-for-in rule to use walk function

### LINKED ISSUE #680: Convert all existing rules to use a `walk` function instead of extending AbstractRule
Partner to #586: following palantir/tslint#2522, we should be recommending a ts.forEachChild strategy.

Accepting any PRs that take an existing rule and convert it to `walk`! ✨

drexler:
#### PR checklist

-   [x] Addresses an existing issue: fixes #680 

#### Overview of change:

-  Updates the template used for generating new rules to use the `walk` function. 
-  Updates the no-for-in rule to use the walk function.

#680 covers all the rules,  so this is just the first of hopefully many...bandwidth permitting. 😄


src/noForInRule.ts
```diff
@@ -25,16 +25,21 @@ export class Rule extends Lint.Rules.AbstractRule {
...
     }
 }
 
-class NoForInRuleWalker extends Lint.RuleWalker {
-    protected visitForInStatement(node: ts.ForInStatement): void {
-        const initializer: string = node.initializer.getText();
-        const expression: string = node.expression.getText();
+function walk(ctx: Lint.WalkContext<void>) {
+    function cb(node: ts.Node): void {
+        if (node.kind === ts.SyntaxKind.ForInStatement) {
```
JoshuaKGoldberg: > `node.kind === ts.SyntaxKind.ForInStatement`

Use `tsutils.isForInStatement(node)`. It returns `node is ts.ForInStatement` as a hint to the type checker so `node` will be narrowed to that type in the for loop. That way you won't have to cast `(<ts.ForInStatement>node)` in there.

build-tasks/templates/rule.template.js
```diff
@@ -8,8 +8,6 @@ import {AstUtils} from './utils/AstUtils';
 // use Utils instead of Underscore functions // TODO: delete comment
```
JoshuaKGoldberg: Looks like changes unintentionally brought in from a recent PR; please revert.

_(if you fix the other comment I can go ahead and do this for you - it shouldn't block the PR)_

drexler: I'll fix these up.

build-tasks/templates/rule.template.js
```diff
@@ -3,12 +3,6 @@ module.exports = ({ ruleName, walkerName }) =>
 import * as Lint from 'tslint';
 
 import {ExtendedMetadata} from './utils/ExtendedMetadata';
-// use (and contribute to) AstUtils for common AST functions // TODO: delete comment
-import {AstUtils} from './utils/AstUtils';
-// use Utils instead of Underscore functions // TODO: delete comment
-import {Utils} from './utils/Utils';
-
```
drexler: Removed the imports as well as they were unused.

build-tasks/templates/rule.template.js
JoshuaKGoldberg: Sorry, I wasn't very clear - no changes to this file should be in this PR (to keep PRs small and single-purpose). If you'd like to send them in, it should be discussed in a separate issue.

build-tasks/templates/rule.template.js
JoshuaKGoldberg: I reverted the changes in this file in a manually made commit so if you're interested in sending them in later, they'll still be here. 👍

drexler:
Thanks for the review @JoshuaKGoldberg

Issue #684, opened 2018-12-18
PR builds on Node 6 fail during linting


JoshuaKGoldberg:
Example PR: https://github.com/Microsoft/tslint-microsoft-contrib/pull/682
Example build break: https://travis-ci.org/Microsoft/tslint-microsoft-contrib/jobs/469577861

```
/home/travis/build/Microsoft/tslint-microsoft-contrib/src/noUnnecessaryOverrideRule.ts:63:51
ERROR: 63:51   increment-decrement  Use an explicit += 1 operator.
/home/travis/build/Microsoft/tslint-microsoft-contrib/src/utils/BaseFormatter.ts:21:62
ERROR: 21:62   increment-decrement  Use an explicit -= 1 operator.
/home/travis/build/Microsoft/tslint-microsoft-contrib/src/utils/Utils.ts:13:46
ERROR: 13:46   increment-decrement  Use an explicit += 1 operator.
/home/travis/build/Microsoft/tslint-microsoft-contrib/src/validTypeofRule.ts:97:40
ERROR: 97:40   increment-decrement  Use an explicit += 1 operator.
ERROR: 102:40  increment-decrement  Use an explicit += 1 operator.
ERROR: 107:40  increment-decrement  Use an explicit += 1 operator.
ERROR: 108:44  increment-decrement  Use an explicit += 1 operator.
```

JoshuaKGoldberg:
`increment-decrement` was added in the just-released TSLint 5.12.0. It's spooky and mysterious that Node 6 isn't respecting `package-lock.json`. This should be fixed by updating the dependency version.

PR #688, opened 2018-12-19
convert mocha-avoid-only rule to use walk function

### LINKED ISSUE #680: Convert all existing rules to use a `walk` function instead of extending AbstractRule

drexler:
#### PR checklist

-   [x] Addresses an existing issue: fixes #680 

#### Overview of change:
Updates the mocha-avoid-only rule to use the walk function.

### DIFF
--- a/src/mochaAvoidOnlyRule.ts
+++ b/src/mochaAvoidOnlyRule.ts
@@ -1,5 +1,6 @@
 import * as ts from 'typescript';
 import * as Lint from 'tslint';
+import * as tsutils from 'tsutils';
 
 import { ExtendedMetadata } from './utils/ExtendedMetadata';
 import { MochaUtils } from './utils/MochaUtils';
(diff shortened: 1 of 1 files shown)


src/mochaAvoidOnlyRule.ts
```diff
@@ -25,38 +26,46 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-                            this.addFailureAt(node.getStart(), node.expression.getText().length, Rule.FAILURE_STRING_SPECIFY);
-                        } else if (node.expression.getText() === 'describe.only') {
-                            this.addFailureAt(node.getStart(), node.expression.getText().length, Rule.FAILURE_STRING_DESCRIBE);
-                        } else if (node.expression.getText() === 'context.only') {
-                            this.addFailureAt(node.getStart(), node.expression.getText().length, Rule.FAILURE_STRING_CONTEXT);
+                                case 'context.only':
+                                    ctx.addFailureAt(node.getStart(), text.length, Rule.FAILURE_STRING_CONTEXT);
+                                    break;
+
+                                default: // required per tslint rules for switch statements.
```
drexler: Intentionally went with the `switch` statement over the `if/elseif` for clarity. The forced inclusion of `default` mars it, admittedly. I can switch to the `if/else if` if required.

drexler:
build failure due to #684 😞

src/mochaAvoidOnlyRule.ts
JoshuaKGoldberg: +1 for getting rid of continuous `.getText()` calls!

JoshuaKGoldberg:
Build failures are pre-existing, unrelated and only on Node 6. #684

PR #692, opened 2018-12-21
Implement react-a11y-iframes rule

### LINKED ISSUE #204: new rule: react-a11y-iframe
All iframe elements must have a title: 
    • https://www.w3.org/TR/WCAG20-TECHS/H64.html
    • http://oaa-accessibility.org/wcag20/rule/10/

Title attributes for frames must be unique.
    • http://oaa-accessibility.org/wcag20/rule/11/

Frames should not be hidden or empty.
    • http://oaa-accessibility.org/wcag20/rule/12/
Reference: https://www.w3.org/TR/WCAG10/wai-pageauth.html#tech-frame-longdesc

Notes: 
    • frame/iframe longdesc attribute is deprecated and not supported by React. 
frame element not supported by React

noamyogev84:
#### PR checklist

-   [x] Addresses an existing issue: fixes #204 
-   [x] enhancement
-   [x] Includes tests
-   [x] Documentation update

#### Overview of change:
 - Validate that `title` attribute exists, and is a non-empty expression or text.
 - Validate that `src` attribute exists, and is a non-empty text.
 - Validate that `hidden` attribute doesn't exist.
 - Validate that `title` attribute text is unique when there are 2 or more `iframe` elements in scope.


noamyogev84:
Hi @JoshuaKGoldberg. 
I'd love your review on this one. Please let me know if you think more scenarios (React Components, Functions) should be covered as well.

IllusionMH:
Hi @noamyogev84 ! Thanks for rule implementation!

We are in process of migration to TSLint test format now (see #489).
Could you convert tests to the TSLint format? If not, it's OK and we will convert it later.

noamyogev84:
@illusionMH sure!

noamyogev84:
Hi @IllusionMH  no problem, my pleasure!
Trying to play a little bit with the new test format. Running ``node --inspect-brk ./node_modules/tslint/bin/tslint --test -r dist/src "tests/**"/**"`` and no breakpoints are being hit in chrome devTools. Am i missing something?

IllusionMH:
There are prepared debug configs if you use VS Code.

I had problem with adding folder to DevTools (in Opera) so
I've added at the beginning of the file
```
// tslint:disable-next-line
debugger;
```
to get to file, and then was able to place breakpoints where I like (there should be original mapped TS code).


If you are on Windows, it might be useful to replace `"sourceMap": true,` with `"inlineSourceMap": true,'` in `tsconfig.json` if approach above doesn't work for you

noamyogev84:
No success on both approaches.
It seems like the vscode debugger just can't find anything to debug with the `tslint` configuration. :(

noamyogev84:
Not working for any rule.
pasted code does contain extra **, fixed it and it's still doesn't work.

README.md
```diff
@@ -151,6 +151,15 @@ We recommend you specify exact versions of lint libraries, including `tslint-mic
       </td>
       <td>6.0.0</td>
     </tr>
+    <tr>
+      <td>
+        <code>react-a11y-iframes</code>
+      </td>
+      <td>
+        Enforce that iframe elements are not empty, have title and are unique..
```
JoshuaKGoldberg: > `are unique..`

Nit: two periods

src/reactA11yIframesRule.ts
```diff
@@ -0,0 +1,99 @@
+import * as ts from 'typescript';
+import * as Lint from 'tslint';
+import * as tsUtils from 'tsutils';
```
JoshuaKGoldberg: > `tsUtils`

Nit: it's normally referred to as `tsutils`, could you please lowercase?

src/reactA11yIframesRule.ts
```diff
@@ -0,0 +1,99 @@
...
+const HIDDEN_ATTRIBUTE_NAME: string = 'hidden';
+const IFRAME_EMPTY_TITLE_ERROR_STRING: string = 'An iframe element must have a non-empty title.';
+const IFRAME_EMPTY_OR_HIDDEN_ERROR_STRING: string = 'An iframe element should not be hidden or empty.';
+const IFRAME_UNIQUE_TITLE_ERROR_STRING: string = 'An iframe element must have a unique title.';
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static metadata: ExtendedMetadata = {
+        ruleName: 'react-a11y-iframes',
+        type: 'functionality',
+        description: 'Enforce that iframe elements are not empty, have title and are unique.',
```
JoshuaKGoldberg: > have title and are

Nit: a [serial comma](https://en.wikipedia.org/wiki/Serial_comma) should be after 'title'. Here and in the README.md please.

src/reactA11yIframesRule.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return sourceFile.languageVariant === ts.LanguageVariant.JSX
+            ? this.applyWithWalker(new ReactA11yIframesRuleWalker(sourceFile, this.getOptions()))
+            : [];
+    }
+}
+
+class ReactA11yIframesRuleWalker extends Lint.RuleWalker {
+    protected visitVariableDeclaration(node: ts.VariableDeclaration): void {
```
JoshuaKGoldberg: Missing a `super.visitVariableDeclaration(node);`. Without that, the `RuleWalker` won't recurse into child nodes.

IllusionMH:
> Will leave testing changes to IllusionMH :)

Will do!

**noamyogev84** if you still have problems with debug (can't find last message here) - could you please  create issue node version and system info. Will investigate it further.

JoshuaKGoldberg:
Oh @IllusionMH I just noticed the ambiguity in my statement - I meant leaving reviewing the changes to the rule's tests to you, not leaving testing out the rule. 😄

README.md
```diff
@@ -151,6 +151,15 @@ We recommend you specify exact versions of lint libraries, including `tslint-mic
...
+      </td>
+      <td>6.0.0</td>
```
IllusionMH: `6.0.0` is already released so this rule will be introduced in `6.1.0`

src/reactA11yIframesRule.ts
noamyogev84: on it :books:

noamyogev84:
@JoshuaKGoldberg @IllusionMH maybe some additional input: Tests for `React` related code must have a `tsx` extension to set the `languageVariant` properly. the following code explains it best: 
```ts
return sourceFile.languageVariant === ts.LanguageVariant.JSX
            ? this.applyWithWalker(new ReactA11yIframesRuleWalker(sourceFile, this.getOptions()))
            : [];```

**Comment by IllusionMH** [comment]
@noamyogev84 sorry, I don't actually understand your last comment.

If you ask about how file with JSX tags should be named in new tests format then it should be `test.tsx.lint` (you can also check [tests in tslint-react repo](https://github.com/palantir/tslint-react/tree/master/test/rules))

If something else - clarify it, please.

**Comment by noamyogev84** [comment]
@IllusionMH yeah well it seems pretty obvious i guess, i wasn't familiar with that. 😞 
Anyway, tests are now included for you to review. 🥂

**Comment by IllusionMH** [comment]
@noamyogev84 thanks! I will review it tonight.

File: src/reactA11yIframesRule.ts (hunk 3 above)
Comment (IllusionMH) [reply src/reactA11yIframesRule.ts]: Looks like this one is not changed.
@JoshuaKGoldberg do I understand correctly that `,` is required after `title` here?

**Comment by noamyogev84** [comment]
I'll start by re implementing the rule with the `walk` function.
@JoshuaKGoldberg @IllusionMH would you care to explain what's the reason we're changing all the rules to this pattern? pretty interesting to understand 😄

**Comment by IllusionMH** [comment]
TL;DR; `walk` function approach easier to control (e.g. reset counters/maps after children iteration) and should be faster (especially if early stop walking through nodes that definitely won't have an error earlier).

There is an issue in TSLint to [mark RuleWalker as deprecated](https://github.com/palantir/tslint/issues/4298) (you can also check [linked discussion](https://github.com/palantir/tslint/issues/2522)).

I think it is similar to `tsUtils.forEachToken`, but with greater control on each step, instead of simple callbacks (as I see there is no way to stop further iteration there).

File: src/reactA11yIframesRule.ts [hunk 6]
```diff
@@ -0,0 +1,87 @@
...
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return sourceFile.languageVariant === ts.LanguageVariant.JSX ? this.applyWithFunction(sourceFile, walk) : [];
+    }
+}
+
+function walk(ctx: Lint.WalkContext<void>) {
+    const previousTitles: Set<string> = new Set();
+    function cb(node: ts.Node): void {
+        if (
+            tsutils.isJsxFragment(node) ||
```
IllusionMH: Not sure that it makes sense to reset titles list when `JsxFragment` is encountered.
`<>` can be between two `iframe`s with identical titles - which will lead to false-negative result.

If titles list should be reset on `<>` I think then special checks for `<React.Fragment>` and `<Fragment>` (depending on import in ideal case) should be added for symmetry.

What do you think?

tests/react-a11y-iframes/test.tsx.lint
```diff
@@ -0,0 +1,63 @@
+
+const SomeComponent = () => <iframe title="I'm a non empty title" src="http://someSource.com"/>;
+
+const SomeComponent = () => <iframe title={"hello there"} src="http://someSource.com"/>;
```
JoshuaKGoldberg: Missing test case: `<iframe title={someVariable} ... />`

noamyogev84: @JoshuaKGoldberg added it quickly.
Just to clarify, you don't mean that this rule should check what `someVar` evaluates to, right?

JoshuaKGoldberg: Correct, I think that's a bit out of scope. Though it might be doable 😉 this seems great as is.

JoshuaKGoldberg:
Whoo, thanks @noamyogev84 and @IllusionMH! 🙌

PR #698, opened 2018-12-27
Removed TSLint@5.12.0 rules from the recommended ruleset

### LINKED ISSUE #697: Remove TSLint@5.12.0 rules from the recommended ruleset (before the next release)
Following #686 - rules from TSLint@5.12.0 are now in the recommended ruleset. That'd be a breaking change and IMO we shouldn't release a minor version before taking them back out.

This is a precursor to, and the opposite of #694 - that one is for adding them back in as a breaking change.

JoshuaKGoldberg:
#### PR checklist

-   [x] Addresses an existing issue: fixes #697

#### Overview of change:

Adding those rules is a breaking change, so we don't want it to come in unplanned.


build-tasks/common/meta.js
```diff
@@ -2,11 +2,24 @@ const path = require('path');
 const glob = require('glob');
 const { readJSON } = require('./files');
 
+/**
+ * @see https://github.com/Microsoft/tslint-microsoft-contrib/issues/694
+ */
+const ignoredRulesNotYetAdded = ['banTsIgnoreRule.js', 'commentTypeRule.js', 'noDefaultImportRule.js', 'unnecessaryConstructorRule.js'];
```
IllusionMH: With exclusion list in this function - new TSLint rules will also be excluded for `rule-metadata.json` and `tslint-warnings.csv` (this change is missing from this PR), however as I understand - main reason is to avoid changes in `recommended_ruleset.js` (and then package `tslint.json`).

Therefore I think it will be better to list of excluded rules to `generate-recommendations.js` or separate file (if more minor updated are possible before `7.0`). This will exclude rules from recommendation, but `tslint-warnings.csv` will still list them as recommended (or not if specified in `additional_rule_metadata.json`). It will also allow checks by rule name instead of file names.

build-tasks/generate-recommendations.js
```diff
@@ -8,30 +8,41 @@ const { getAllRules, getMetadataFromFile, getMetadataValue } = require('./common
...
+    )
+);
+
+getAllRules()
+    .filter(ruleFilePath => !ignoredRulesNotYetAdded.has(ruleFilePath))
+    .forEach(ruleFile => {
+        const metadata = getMetadataFromFile(ruleFile);
+
+        const groupName = getMetadataValue(metadata, 'group');
+        if (groupName === 'Ignored') {
```
IllusionMH: For simplicity this condition can be extended to also check if `ruleName` (need to be moved from line 43 up) is in `ignoredRulesNotYetAdded` that can be set of rule names instead of file paths.

Will remove `.filter` on all rules and `.map` for set creation.

IllusionMH:
And one more question (unable to comment inline):
`comment-type` is intentionally left as not recommended ([`"recommendation": "false,"`](https://github.com/Microsoft/tslint-microsoft-contrib/pull/698/files#diff-88ec5f17177774ca2c7ad2058e5f652aR58))?
It doesn't affect package `tslint.json`, but marked as not recommended in `tslint-warnings.csv`

JoshuaKGoldberg:
> `comment-type`

Should be excluded from recommendations now.

additional_rule_metadata.json
```diff
@@ -28,7 +28,6 @@
     "ban-ts-ignore": {
```
JoshuaKGoldberg: Will remove the remaining recommendations here...

IllusionMH:
Looks great!

PR #704, opened 2018-12-28
convert no-constant-condition rule to use walk function

drexler:
#### PR checklist

#### Overview of change:
Converts `no-constant-condition` rule to use walk function

### DIFF
--- a/src/noConstantConditionRule.ts
+++ b/src/noConstantConditionRule.ts
@@ -1,9 +1,14 @@
 import * as ts from 'typescript';
 import * as Lint from 'tslint';
+import * as tsutils from 'tsutils';
 
 import { AstUtils } from './utils/AstUtils';
 import { ExtendedMetadata } from './utils/ExtendedMetadata';
 import { isObject } from './utils/TypeGuard';
+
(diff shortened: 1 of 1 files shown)


src/noConstantConditionRule.ts
```diff
@@ -24,73 +29,63 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+
+function walk(ctx: Lint.WalkContext<Options>) {
+    const { checkLoops } = ctx.options;
 
-    protected visitWhileStatement(node: ts.WhileStatement): void {
-        if (this.checkLoops) {
+    function cb(node: ts.Node): void {
+        if (checkLoops && (tsutils.isWhileStatement(node) || tsutils.isDoStatement(node))) {
             if (AstUtils.isConstantExpression(node.expression)) {
                 const message: string = Rule.FAILURE_STRING + 'while (' + node.expression.getText() + ')';
```
JoshuaKGoldberg: > `'while ('`

Huh, looks like an existing bug with the rule. I'll file a separate issue; not something that should be addressed in this PR.

PR #720, opened 2018-12-30
Converted react-unused-props-and-state to use a walk function.

reduckted:
#### PR checklist

-   ~~New feature, bugfix, or enhancement~~
    -   ~~Includes tests~~
-   ~~Documentation update~~

#### Overview of change:

Converted react-unused-props-and-state to use a walk function.

This one's a bit odd because it sort of does two passes over the AST. It walks everything once, but _doesn't_ step down on `ClassDeclaration` nodes. Then it does a "second pass" by stepping down on the `ClassDeclaration` nodes it saw in the first pass. I considered splitting it up into two separate callback functions, but I wasn't sure exactly what's expected to be visited on the first pass and what's visited on the second pass, so I just used one callback function which is the equivalent of what the walker class did.

### DIFF
--- a/src/reactUnusedPropsAndStateRule.ts
+++ b/src/reactUnusedPropsAndStateRule.ts
@@ -1,5 +1,6 @@
 import * as ts from 'typescript';
 import * as Lint from 'tslint';
+import * as tsutils from 'tsutils';
 
 import { Utils } from './utils/Utils';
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg:
Thanks for the notes - seems fine the way you have it. This is kind of a spooky rule.

PR #721, opened 2018-12-30
Converted react-no-dangerous-html to use a walk function.

reduckted:
#### PR checklist

-   ~~New feature, bugfix, or enhancement~~
    -   ~~Includes tests~~
-   ~~Documentation update~~

#### Overview of change:

Converted react-no-dangerous-html to use a walk function.

The public static `Rule.getExceptions()` function was supposedly "Exposed for testing.", but nothing outside of this class used it, so I've removed it. I'm not sure if that's considered a breaking change (it was a public function on an exported class). If it is a breaking change, then I can easily revert it by rolling back the second commit.


JoshuaKGoldberg:
Rules aren't generally considered part of TSLint's API, so I don't think we should consider this a breaking change. ✔

PR #754, opened 2019-01-01
exclude specific input types from react-a11y-input-elements rule

### LINKED ISSUE #749: Rule `react-a11y-input-elements` incorrectly requiring placeholder for radio, checkbox, and file inputs
### Bug Report

-   **`tslint-microsoft-contrib` version**: 6.0.0
-   **TSLint version**: 5.12.0
-   **TypeScript version**: 3.2.2
-   **Running TSLint via**: CLI

```tsx
<input
    type="checkbox"
    checked={props.isSelected}
    onChange={onChange}
/>
```

```json

  "extends": [
    "tslint:latest",
    "tslint-microsoft-contrib/recommended_ruleset"
  ],

  "rulesDirectory": [
    "tslint-microsoft-contrib"
  ],

  "rules": {
    "quotemark": [true, "single", "jsx-double"],
  }
}
```

#### Actual behavior

ERROR: 40:17    react-a11y-input-elements  Input elements must include default, place-holding characters if empty

#### Expected behavior

Inputs with type checkbox, radio, and, file should not require a placeholder value.

lizzzp1:
#### PR checklist

-   [x] Addresses an existing issue: fixes #749
    -   [x] Includes tests
-   [ ] Documentation update

#### Overview of change:
Excludes the 3 input types listed in issue from requiring a placeholder.

### DIFF
--- a/src/reactA11yInputElementsRule.ts
+++ b/src/reactA11yInputElementsRule.ts
@@ -8,6 +8,7 @@ import { ExtendedMetadata } from './utils/ExtendedMetadata';
 export const MISSING_PLACEHOLDER_INPUT_FAILURE_STRING: string = 'Input elements must include default, place-holding characters if empty';
 export const MISSING_PLACEHOLDER_TEXTAREA_FAILURE_STRING: string =
     'Textarea elements must include default, place-holding characters if empty';
+const EXCLUDED_INPUT_TYPES = ['checkbox', 'radio', 'file'];
 
 /**
  * Implementation of the react-a11y-input-elements rule.
@@ -45,17 +46,34 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 }
 
+function isExcludedInputType(node: ts.JsxSelfClosingElement): boolean {
+    let isExcludedType = false;
+    node.attributes.properties.forEach(
+        (attribute: ts.JsxAttributeLike): void => {
+            if (attribute.kind === ts.SyntaxKind.JsxAttribute) {
+                if (attribute.initializer !== undefined && attribute.initializer.kind === ts.SyntaxKind.JsxExpression) {
+                    const attributeText: string = (<ts.StringLiteral>(<ts.JsxAttribute>attribute).initializer).text;
+                    if (EXCLUDED_INPUT_TYPES.indexOf(attributeText) !== -1) {
+                        isExcludedType = true;
+                    }
+                }
+            }
+        }
+    );
+    return isExcludedType;
+}
+
 function walk(ctx: Lint.WalkContext<void>) {
     function cb(node: ts.Node): void {
         if (tsutils.isJsxSelfClosingElement(node)) {
             const tagName = node.tagName.getText();
 
-            if (tagName === 'input') {
+            if (tagName === 'input' && !isExcludedInputType(node)) {
                 const attributes = getJsxAttributesFromJsxElement(node);
                 if (isEmpty(attributes.value) && isEmpty(attributes.placeholder)) {
                     ctx.addFailureAt(node.getStart(), node.getWidth(), MISSING_PLACEHOLDER_INPUT_FAILURE_STRING);
                 }
-            } else if (tagName === 'textarea') {
+            } else if (tagName === 'textarea' && !isExcludedInputType(node)) {
                 const attributes = getJsxAttributesFromJsxElement(node);
                 if (isEmpty(attributes.placeholder)) {
                     ctx.addFailureAt(node.getStart(), node.getWidth(), MISSING_PLACEHOLDER_TEXTAREA_FAILURE_STRING);
--- a/src/reactA11yInputElementsRule.ts
+++ b/src/reactA11yInputElementsRule.ts
@@ -48,18 +48,18 @@ export class Rule extends Lint.Rules.AbstractRule {
 
 function isExcludedInputType(node: ts.JsxSelfClosingElement): boolean {
     let isExcludedType = false;
-    node.attributes.properties.forEach(
-        (attribute: ts.JsxAttributeLike): void => {
-            if (attribute.kind === ts.SyntaxKind.JsxAttribute) {
-                if (attribute.initializer !== undefined && attribute.initializer.kind === ts.SyntaxKind.JsxExpression) {
-                    const attributeText: string = (<ts.StringLiteral>(<ts.JsxAttribute>attribute).initializer).text;
-                    if (EXCLUDED_INPUT_TYPES.indexOf(attributeText) !== -1) {
-                        isExcludedType = true;
-                    }
+
+    for (const attribute of node.attributes.properties) {
+        if (tsutils.isJsxAttribute(attribute)) {
+            if (attribute.initializer !== undefined && tsutils.isStringLiteral(attribute.initializer)) {
+                const attributeText = attribute.initializer.text;
+                if (EXCLUDED_INPUT_TYPES.indexOf(attributeText) !== -1) {
+                    isExcludedType = true;
+                    return isExcludedType;
                 }
             }
         }
-    );
+    }
     return isExcludedType;
 }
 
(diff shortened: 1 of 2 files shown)


src/reactA11yInputElementsRule.ts
```diff
@@ -45,17 +46,34 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 }
 
+function isExcludedInputType(node: ts.JsxSelfClosingElement): boolean {
+    let isExcludedType = false;
+    node.attributes.properties.forEach(
+        (attribute: ts.JsxAttributeLike): void => {
+            if (attribute.kind === ts.SyntaxKind.JsxAttribute) {
```
JoshuaKGoldberg: Use `tsutils.isJsxAttribute` on `attribute` and `tsutils.isStringLiteral` on `attribute.initializer. They return whether a given node is of those types, so you won't have to `<>` cast them later on. Nifty hints to the type system.

src/tests/ReactA11yInputElementsRuleTests.ts
```diff
@@ -20,10 +20,21 @@ describe('reactA11yInputElementsRule', (): void => {
         TestHelper.assertViolations(ruleName, script, []);
     });
 
+    it('should pass on input elements without placeholder of type radio, checkbox, file', (): void => {
```
JoshuaKGoldberg: > should pass

Should fail, right? Looks like your change works (🎉) but the test is failing.

src/reactA11yInputElementsRule.ts
```diff
@@ -45,17 +46,34 @@ export class Rule extends Lint.Rules.AbstractRule {
...
             const tagName = node.tagName.getText();
 
-            if (tagName === 'input') {
+            if (tagName === 'input' && !isExcludedInputType(node)) {
                 const attributes = getJsxAttributesFromJsxElement(node);
                 if (isEmpty(attributes.value) && isEmpty(attributes.placeholder)) {
                     ctx.addFailureAt(node.getStart(), node.getWidth(), MISSING_PLACEHOLDER_INPUT_FAILURE_STRING);
                 }
-            } else if (tagName === 'textarea') {
+            } else if (tagName === 'textarea' && !isExcludedInputType(node)) {
```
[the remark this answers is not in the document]
lizzzp1: Oh oops! You're right

src/reactA11yInputElementsRule.ts
```diff
@@ -45,17 +46,34 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 }
 
+function isExcludedInputType(node: ts.JsxSelfClosingElement): boolean {
+    let isExcludedType = false;
+    node.attributes.properties.forEach(
```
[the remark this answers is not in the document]
lizzzp1: ah, yes. Good point.

src/reactA11yInputElementsRule.ts
lizzzp1: Ah. Thank you! I was digging in the TS source code for these types and couldn't find this one.

src/tests/ReactA11yInputElementsRuleTests.ts
JoshuaKGoldberg: Oh sorry you're right, I had it mixed up. The test is right but the code looks like it has a bug.

src/reactA11yInputElementsRule.ts
```diff
@@ -45,17 +46,34 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        (attribute: ts.JsxAttributeLike): void => {
+            if (attribute.kind === ts.SyntaxKind.JsxAttribute) {
+                if (attribute.initializer !== undefined && attribute.initializer.kind === ts.SyntaxKind.JsxExpression) {
```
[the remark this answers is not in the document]
JoshuaKGoldberg: > checking properties

I believe so? It should be picking up any attribute on the element, so you'll want to only look at an attribute if it actually is of name `type`

> `isEmpty`

Yes. If a blank type is provided it should default to text IIRC.

src/reactA11yInputElementsRule.ts
```diff
@@ -45,12 +46,29 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 }
 
+function isExcludedInputType(node: ts.JsxSelfClosingElement): boolean {
+    let isExcludedType = false;
```
JoshuaKGoldberg: Not: since this variable is only ever assigned to once, immediately before a return, let's remove it and directly return `true` or `false`. A little less code.

src/tests/ReactA11yInputElementsRuleTests.ts
```diff
@@ -20,10 +20,21 @@ describe('reactA11yInputElementsRule', (): void => {
...
+        const script: string = `
+            import React = require('react');
```
JoshuaKGoldberg: Could you please also add tests for the case of inputs with attribute names other than `type`? Tests are passing without that check in the rule, so they're missing that case.

lizzzp1: Will do!

src/reactA11yInputElementsRule.ts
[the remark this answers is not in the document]
lizzzp1: So if it's an excluded type and there is no value attribute provided should it give you a more specific failure string? Right now it appears to give you the same failure error when value or placeholder isn't provided.
`Input elements must include default, place-holding characters if empty` Would this make sense if you don't provide a value?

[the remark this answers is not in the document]
lizzzp1: Sure, sounds good.

src/reactA11yInputElementsRule.ts
```diff
@@ -47,30 +47,30 @@ export class Rule extends Lint.Rules.AbstractRule {
...
             const tagName = node.tagName.getText();
 
-            if (tagName === 'input' && !isExcludedInputType(node)) {
+            if (tagName === 'input') {
                 const attributes = getJsxAttributesFromJsxElement(node);
-                if (isEmpty(attributes.value) && isEmpty(attributes.placeholder)) {
+                const isExcludedInputTypeValueEmpty = isEmpty(attributes.value) && isExcludedInputType(node);
+                const isPlaceholderEmpty = isEmpty(attributes.placeholder) && !isExcludedInputType(node);
+                if ((isEmpty(attributes.value) && isPlaceholderEmpty) || isExcludedInputTypeValueEmpty) {
                     ctx.addFailureAt(node.getStart(), node.getWidth(), MISSING_PLACEHOLDER_INPUT_FAILURE_STRING);
```
lizzzp1: 😬 might this be simplified?

JoshuaKGoldberg: A couple things to try:

* Pass `attributes` into `isExcludedInputType.type` so you don't have to call `getJsxAttributesFromJsxElement` each time
* Only make one call to `isExcludedInputType(node)` and store that in a variable, so you don't need to call it twice.

...but the test cases look good, build passes, and this is still a pretty simple rule; I'm fine with merging this in as is & filing a followup issue to look into cleaning it up. 

I'll wait for direction from you if you want to tackle it now - totally up to you! 😊

src/reactA11yInputElementsRule.ts
```diff
@@ -46,10 +46,10 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 }
 
-function isExcludedInputType(node: ts.JsxSelfClosingElement): boolean {
+function isExcludedInputType(node: ts.JsxSelfClosingElement, attributes: { [propName: string]: ts.JsxAttribute }): boolean {
     for (const attribute of node.attributes.properties) {
```
lizzzp1: got this type from below. Let me know if this looks right to you.

src/reactA11yInputElementsRule.ts
lizzzp1: Good call! Dried it up ⬇️

src/reactA11yInputElementsRule.ts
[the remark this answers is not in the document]
lizzzp1: What type of issue does this fall under?

PR #665, opened 2018-12-04
Converted open rules from a c* to no-*

JoshuaKGoldberg:
#### PR checklist

-   [x] Addresses an existing issue: continues #489

#### Overview of change:

Excluding @IllusionMH's 🏃 rules, converts list items mentioned starting from `chai-*` through `no-*` rules.

PR #770, opened 2019-01-07
void-0-rule

### LINKED ISSUE #693: Rule suggestion: void-zero
### Rule Suggestion

**Is your rule for a general problem or is it specific to your development style?**

One could argue for either, but IMO this is more of a general problem.

**What does your suggested rule do?**

Some projects have instances of `void 0` instead of `undefined`. They technically resolve to the same thing _(and minifiers typically convert to `void 0` to safe a few characters)_ but `void 0` can be confusing to newcomers. As a stylistic preference to make it easier for newcomers I think it'd be reasonable to have a rule to prefer `undefined`.

**List several examples where your rule could be used**

[VS Code](https://github.com/microsoft/vscode) has a lot of these.

Given:

```typescript
function takesMaybe(param1: string | undefined, param2: number) { /* ... */ }
```

Instead of:

```typescript
function takesMaybe(void 0, 7);
```

Prefer:

```typescript
function takesMaybe(undefined, 7);
```

**Additional context**

/cc @roblourens because I get a little irked every time I wade into VS Code source 😄

lizzzp1:
#### PR checklist

-   [x] Addresses an existing issue: fixes #693
    -   [x] Includes tests
-   [ ] Documentation update

#### Overview of change:
Attempts to add a no void 0 rule.

@JoshuaKGoldberg unsure if this is the right implementation. It seems to evaluate void and zero individually as parameters (the test confirms it gives 2 errors). Not sure if looking for the specific string in sourceFile is brittle. Also.. I may have missed some cases?


src/tests/NoVoidZeroRuleTests.ts
```diff
@@ -0,0 +1,62 @@
+import { Utils } from '../utils/Utils';
```
JoshuaKGoldberg: Would you mind switching this to the newer test format, actually? Under the `test/` directory, copy any one that doesn't have sub-directories _(which would be necessary if this rule has options)_ such as `non-literal-fs-path`, modify the `tslint.json` to only include `no-void-0`, and add your test cases there. 

Tests in general are moving to that format (#489) because it's quite a bit nicer to read & write.

src/noVoidZeroRule.ts
```diff
@@ -0,0 +1,61 @@
...
+import * as Lint from 'tslint';
+import * as tsutils from 'tsutils';
+
+import { ExtendedMetadata } from './utils/ExtendedMetadata';
+
+const FAILURE_STRING: string = 'Replace void 0 with undefined';
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static metadata: ExtendedMetadata = {
+        ruleName: 'no-void-zero',
```
JoshuaKGoldberg: > `ruleName: 'no-void-zero',

IMO it's better to move away from prefixing rule names with `no-` for a couple of reasons:

* It discourages rules from later being modified to support the opposite case. That's not so relevant to this rule, but does for others _(e.g. `parameter-reassignment`; what if we want prefer reassigning them over making unnecessary new variables?)_
* It messes with sorting of rules, since some have `no-` and others don't

JoshuaKGoldberg:
Re your edge cases - I can't think of too many others right now 😄 but it seems fine to go in with just a few. Maybe the different ways variables/properties can be assigned, such as parameters, variables, class members, and static class members?

src/noVoidZeroRule.ts
```diff
@@ -0,0 +1,61 @@
+import * as ts from 'typescript';
+import * as Lint from 'tslint';
+import * as tsutils from 'tsutils';
+
+import { ExtendedMetadata } from './utils/ExtendedMetadata';
+
+const FAILURE_STRING: string = 'Replace void 0 with undefined';
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static metadata: ExtendedMetadata = {
```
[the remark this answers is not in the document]
lizzzp1: Truth. 👍

src/noVoidZeroRule.ts
```diff
@@ -0,0 +1,61 @@
...
+                ctx.addFailureAt(node.getStart(), node.getWidth(), FAILURE_STRING);
+            }
+        }
+        if (
+            tsutils.isMethodDeclaration(node) ||
+            tsutils.isConstructorDeclaration(node) ||
+            tsutils.isArrowFunction(node) ||
+            tsutils.isFunctionDeclaration(node) ||
+            tsutils.isFunctionExpression(node)
+        ) {
```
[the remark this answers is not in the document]
lizzzp1: `Void expressions can't be function parameters` gotcha -- thought I saw it used as so. Makes sense.

src/tests/NoVoidZeroRuleTests.ts
lizzzp1: Sure! Good to know -- thank you!

src/noVoidZeroRule.ts
lizzzp1: 👍 Makes total sense

src/tests/NoVoidZeroRuleTests.ts
```diff
@@ -0,0 +1,62 @@
...
+                ruleName,
+                startPosition: { line: 1, character: 13 }
+            }
+        ]);
+    });
+
+    it('should fail when function is given argument of void 0', (): void => {
+        const inputScript: string = `
+            const baz = (void 0) => {};
+            function (void 0) { }
```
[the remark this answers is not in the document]
lizzzp1: 🙃 😬

src/noVoidZeroRule.ts
lizzzp1: Do you think just `void-zero` is sufficient or `prefer-undefined`?

src/tests/NoVoidZeroRuleTests.ts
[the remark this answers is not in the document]
lizzzp1: `function takesMaybe(void 0, 7)` Ohhh I was looking at this example and thought it was a declaration. duh. I see my misunderstanding now! haha

src/noVoidZeroRule.ts
JoshuaKGoldberg: `void-zero`. names around preferring undefined could get confused with rules in the `null` vs. `undefined` area.

lizzzp1:
Updated. The tests are an interesting format. Would there be no need need for a test runner if these were all this format or would you still need mocha/chai set up?

IllusionMH:
> Updated. The tests are an interesting format. Would there be no need need for a test runner if these were all this format or would you still need mocha/chai set up?

There are tests for utility methods that are used in rules implementation, so mocha/chai setup will be still used, but only those utility methods.

README.md
```diff
@@ -722,6 +722,15 @@ We recommend you specify exact versions of lint libraries, including `tslint-mic
...
       <td>4.0.2</td>
     </tr>
+    <tr>
+      <td>
+        <code>void-zero</code>
+      </td>
+      <td>
+        <code>void 0</code>, which resolves to <code>undefined</code>, can be confusing newcomers. Exclusively use <code>undefined</code> to reduce ambiguity.
+      </td>
+      <td>6.0.0</td>
```
IllusionMH: `6.0.0` is current verstion. This rule will be added in next minor version - `6.1.0`.

src/voidZeroRule.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const FAILURE_STRING: string = 'Replace void 0 with undefined';
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static metadata: ExtendedMetadata = {
+        ruleName: 'void-zero',
+        type: 'maintainability',
+        description: 'Avoid using void 0; use undefined instead.',
+        rationale:
+            'void 0, which resolves to undefined, can be confusing to newcomers.\
```
IllusionMH: Nit: this will produce string with 12 spaces between `.` and `Exclusively`.
<pre>
void 0, which resolves to undefined, can be confusing to newcomers.            Exclusively use undefined to reduce ambiguity.</pre>

Better to just make it single line. Without extra spaces it should fit line limit.

IllusionMH:
This is pretty simple rule with concrete instruction what to do. May be it make sense to do add fixer?

@lizzzp1 would you prefer to add fixer in this PR or should I create separate issue for enhancement?

README.md
lizzzp1: Ah. Thanks

src/voidZeroRule.ts
lizzzp1: Sure!

lizzzp1:
@IllusionMH updated.

README.md
lizzzp1: Oh I thought you were linking to this when you said one more question. Define `fixer` :)

IllusionMH:
_I will post here instead of comment for line_

Fixer is code that will fix lint error (in this case replace `void 0` expression to just `undefind`) if user runs `tslint` with `--fix` option.
Example in this repo: https://github.com/Microsoft/tslint-microsoft-contrib/blob/59fa06e239f0a84da9d55113006ab0e8a6a7553a/src/importNameRule.ts#L175-L176
and 1 line in meta for consistency: https://github.com/Microsoft/tslint-microsoft-contrib/blob/59fa06e239f0a84da9d55113006ab0e8a6a7553a/src/importNameRule.ts#L14

Example in docs (no direct link to paragraph, example in the end): https://palantir.github.io/tslint/develop/custom-rules/

lizzzp1:
@IllusionMH 💡ah, yes the `--fix` flag,  got it. I'm happy to do that here or if you think it's easier to do it in another issues, that's fine too. Would it helpful for me to add this to the contributor read me? This might be common knowledge, but if you think it would be helpful, let me know.

IllusionMH:
> I'm happy to do that here or if you think it's easier to do it in another issues, that's fine too.

@lizzzp1 since fixer for this rule should be easy to implement, I'd prefer to have everything in this PR.

Please note that fixer behavior should be tested with adding `tests/void-zero/test.ts.fix` file that has all replacements in place.
You can check [test.ts.lint](https://github.com/Microsoft/tslint-microsoft-contrib/pull/665/files#diff-e493739808abca8a94169db68735a91b) and corresponding [test.ts.fix](https://github.com/Microsoft/tslint-microsoft-contrib/pull/665/files#diff-dde0e837d97e2876aad8b62498b04320) for example (links to PR).

If you'll ahve any problems do not hesitate to ask here.

>Would it be helpful for me to add this to the contributor read me? This might be common knowledge, but if you think it would be helpful, let me know.

I've created #779 to track this suggestion.

lizzzp1:
@IllusionMH sounds good, thanks!

IllusionMH:
Unfortunately I forgot that #665 is not merged.
You'll need to add `*.fix eol=lf` to `.gitattributes` (see [`.gitattributes` diff](https://github.com/Microsoft/tslint-microsoft-contrib/pull/665/files#diff-fc723d30b02a4cca7a534518111c1a66) to avoid conflicts) and ensure that `tests/void-zero/test.ts.fix` has only `LF` line endings (should have correct already)

JoshuaKGoldberg:
Yeah SGTM! I can work on the other, bigger PRs in a few minutes.

PR #781, opened 2019-01-10
convert missing-jsdoc rule to use a walk function

drexler:
#### PR checklist

#### Overview of change:
Converts `missing-jsdoc` rule to use a walk function

### DIFF
--- a/src/missingJsdocRule.ts
+++ b/src/missingJsdocRule.ts
@@ -1,7 +1,6 @@
 import * as ts from 'typescript';
 import * as Lint from 'tslint';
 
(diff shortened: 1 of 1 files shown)


src/missingJsdocRule.ts
```diff
@@ -1,6 +1,7 @@
 import * as ts from 'typescript';
 import * as Lint from 'tslint';
 
+//import { isSourceFile } from 'tsutils';
```
JoshuaKGoldberg: I'll remove this :)

JoshuaKGoldberg:
Thanks!

PR #792, opened 2019-01-20
convert no-banned-terms & no-reserved-keywords rules to use a walk function

drexler:
#### PR checklist

#### Overview of change:
Converts `no-banned-terms` & `no-reserved-keywords` rules to use a walk function


drexler:
A twofer as the BannedTerms rule walker is utilized across two rules.

PR #794, opened 2019-01-20
add react-a11y-iframes to recommended ruleset

drexler:
#### Overview of change:
Adds react-a11y-iframes to recommended ruleset. This was missing from #692, i reckon.

### DIFF
--- a/recommended_ruleset.js
+++ b/recommended_ruleset.js
@@ -188,6 +188,7 @@ module.exports = {
         'react-a11y-anchors': true,
         'react-a11y-aria-unsupported-elements': true,
         'react-a11y-event-has-role': true,
+        'react-a11y-iframes': true,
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg:
Hmm, I don't know that we _can_ add it to the recommended ruleset. TSLint core doesn't change `recommended` except in major versions (I made that mistake of recommending otherwise with the 5.12.0 release).

drexler:
Interesting. Learnt something new 😄.

IllusionMH:
@drexler thank you for your vigilance. As **JoshuaKGoldberg** mentioned - new rules shouldn't be part of recommended configuration until next major release.

I will close this PR in favor of #809 that creates additional config to include recommended rules in minor versions.

Please let me know if you have any questions.

PR #798, opened 2019-01-27
Changed no-string-based-set-x rules to use a walk function.

reduckted:
#### PR checklist

    -   ~~Includes tests~~
-   ~~Documentation update~~

#### Overview of change:

Changed these rules to use a walk function:
* `no-string-based-set-immediate`
* `no-string-based-set-interval`
* `no-string-based-set-timeout`

All three rules used the same walker class.

Nope.


src/utils/createNoStringParameterToFunctionWalker.ts
```diff
@@ -0,0 +1,210 @@
...
+                scope = new Scope(scope);
+                scope.addParameters(node.parameters);
+                ts.forEachChild(node, cb);
+                scope = scope.parent;
+            } else if (tsutils.isFunctionExpression(node)) {
+                scope = new Scope(scope);
+                scope.addParameters(node.parameters);
+                ts.forEachChild(node, cb);
+                scope = scope.parent;
+            } else if (tsutils.isSetAccessorDeclaration(node)) {
```
JoshuaKGoldberg: A bunch of these conditions are the same. Please switch to a single `if` statement with `||`s for the different cases to reduce code duplication.

src/utils/createNoStringParameterToFunctionWalker.ts
```diff
@@ -0,0 +1,210 @@
...
+                scope = new Scope(scope);
+                scope.addParameters(node.parameters);
+                ts.forEachChild(node, cb);
+                scope = scope.parent;
+            } else {
+                if (tsutils.isVariableDeclaration(node)) {
+                    // this.scope is always set upon entering a source file, so we know it exists here
```
JoshuaKGoldberg: > `this.scope`

Nit: `scope`

Issue #700, opened 2018-12-27
Better document and support auto-generated configs for contributors


JoshuaKGoldberg:
Quoting @IllusionMH in #696:

> I have an impression (may be wrong) that contributors are not sure if they should add/commit changes to automatically generated files (specifically `recommended_ruleset.js` and `tslint-warnings.csv`).

> It may worth to add information to `CONTRIBUTING.md` that changes in these files related to new rules should be also commited. And during review check if changes are present and ask to add for example `recommended: 'false, // to strict'` to rule metadata.

And my response:

> Agreed. I think it'd be good to have a comment on top of `recommended_ruleset.js` linking to those docs too.

I wonder if we should enable them as a commit hook, as with Prettier?

JoshuaKGoldberg:
☠️  *It's time!* ☠️ 

Per #876, this repository is no longer accepting feature pull requests. TSLint is being deprecated and we recommend you switch to https://typescript-eslint.io.

Thanks for open sourcing with us, everyone!

Issue #649, opened 2018-11-07
Don't recommend no-use-before-declare or other deprecated or legacy rules


JoshuaKGoldberg:
Following discussion in https://github.com/vuejs/vue-cli/issues/2712

IllusionMH:
Is there a clear way to determine which rules are "deprecated" or "legacy"?
For example `no-use-before-declare` doesn't have any hints (metadata properties) except couple sentences in description.

In this case - changes should be made to `additional_rule_metadata.json` but this will also affect `tslint-warnings.csv` which I'm not sure how it is used.

JoshuaKGoldberg:
@IllusionMH that's a good idea. You could also consider filing an issue in TSLint core to add "legacy"/"discouraged" field(s) to metadata.

JoshuaKGoldberg:
### 💀 _**It's time!**_ 💀

**TSLint is deprecated** and no longer accepting pull requests other than security fixes. See #876. ☠️ 
We recommend you instead use **[typescript-eslint](https://typescript-eslint.io)** to lint your TypeScript code with **ESLint**. ✅

👋 It was a pleasure open sourcing with you!

PR #809, opened 2019-02-02
Revamped configs into 'core', 'latest', and (legacy) 'recommended'

### LINKED ISSUE #677: Stop extending from TSLint's ruleset in the published ruleset
Followup to #309 - it's confusing to many users that `"extends": ["tslint-microsoft-contrib"]` includes our rules extended from TSLint itself. This should be removed in 7.0.

Community packages should just be a set of rules; there's no reason why extending their rules would have any opinion on the core rules brought in with them.

### LINKED ISSUE #716: Stop including library-specific rules in the recommended ruleset
Spinning off the discussion in #708: enabling the Backbone-related rules in the recommended ruleset is probably just a slight performance tick for the vast majority of users. Can we remove them by default? How about, say, the jQuery rules?

IMO the mocha/chai/etc. test rules should stay in, since that's the general shape of testing APIs in JavaScript test frameworks these days.

JoshuaKGoldberg:
#### PR checklist

-   [x] Addresses an existing issue: starts on #649; fixes #677; starts on #700; fixes #716
-   [x] Documentation update

#### Overview of change:

Those three config files now exist as `.json` under `/configs`; they're copied to `dist/build/configs`. `recommended.json` becomes `tslint.json` as well.

Does this look like a reasonable way to manage configurations for foreseeable future?


VincentLanglet:
@JoshuaKGoldberg The same way tslint expose a `tslint:all` config, you may create a `tslint-microsoft-config:all` config.

JoshuaKGoldberg:
> `:all`

Yes!! Great idea @VincentLanglet, will do.

> `:a11y`

Also strongly in support of this! But - I think it would be better, long-term, to move those rules to tslint-react, as it's been confusing to many users where different contributed rules come from & why. Will file issues there.

VincentLanglet:
> > `:all`
> 
> Yes!! Great idea @VincentLanglet, will do.
> 
> Edit: looking through, it seems like `latest.ts` is already what `:all` would be. Are there other rules/options missing?

I didn't have rules in mind. I was just thinking it was the same naming than `tslint` configs : 

- `recommended` is a subset of all rules, minor version doesn't add breaking changes.
- `latest` is the `recommended` rules for the next major version, but still a `subset` of all rules.
- `all` is all.

If you're going to do: 

- `stable` every rules but without breaking changes for minor version.
- `latest` every rules. Will be the `stable` of the next major version. 

You may rename the `latest` to `all`

> > `:a11y`
> 
> Also strongly in support of this! But - I think it would be better, long-term, to move those rules to tslint-react, as it's been confusing to many users where different contributed rules come from & why. Will file issues there.

It's not a bad idea to move those rules to `tslint-react`. But it's not the same maintainers (`Microsoft` vs `palantir`) ; i don't know the needs and consequences.
Anyway it will be a big breaking change.  The `:a11y` can be a first step.

build-tasks/copy-config-json.js
```diff
@@ -0,0 +1,23 @@
...
+    const resolvedDestination = path.resolve(destinationFileName);
+    const data = readFile(path.resolve(configFileName));
+
+    writeFile(resolvedDestination, data.replace(/"rulesDirectory": \[(.*)\]/, `"rulesDirectory": ["${newRulesDirectory}"]`));
+};
+
+mkdirp.sync('./dist/build/configs');
+
+for (const configFileName of readDirectory('./configs')) {
+    copyConfigFile(`./configs/${configFileName}`, `./dist/build/configs/${configFileName}`, '../');
```
IllusionMH: Is is requirement to nest configurations in `config` folder?
I think that adding configuration to `tslint.json` without `config` will be a bit cleaner, however don't have strong opinion here.

configs/recommended.json
```diff
@@ -0,0 +1,286 @@
...
+        "no-missing-visibility-modifiers": false, // use tslint member-access rule instead
+        "no-multiple-var-decl": false, // use tslint one-variable-per-declaration rule instead
+        "no-reserved-keywords": false,
+        "no-stateless-class": false,
+        "no-switch-case-fall-through": false, // now supported by TypeScript compiler
+        "no-unnecessary-class": true,
+        "no-var-self": false,
+        "react-tsx-curly-spacing": false,
+        "typeof-compare": false, // the valid-typeof rule is currently superior to this version
+        "valid-typeof": false,
```
IllusionMH: Trailling comma can be removed here.

configs/recommended.json
```diff
@@ -0,0 +1,286 @@
...
+    }
+};
```
IllusionMH: No need for `;` in `json` file.

.gitignore
```diff
@@ -1,24 +1,7 @@
...
-# TypeScrit temporary/cache files
-**/.baseDir.ts
-**/.tscache/
-tscommand-*.tmp.txt
-
+# Generated for builds
 **/dist/
-
-# IDEA
-.idea
```
IllusionMH: I think it make sense to leave this folder in ignore list in case any contributor uses WebStorm

configs/stable.json
```diff
@@ -0,0 +1,82 @@
...
+        "react-a11y-titles": true,
+        "react-anchor-blank-noopener": true,
+        "react-iframe-missing-sandbox": true,
+        "react-no-dangerous-html": true,
+        "react-this-binding-issue": true,
+        "react-unused-props-and-state": true,
+        "underscore-consistent-invocation": true,
+        "use-named-parameter": true,
+        "use-simple-attributes": true,
+        "void-zero": true
```
IllusionMH: `void-zero` was added after `6.0.0` and should be in `latest.json`

JoshuaKGoldberg:
Thanks for the feedback you two! I agree with what looks like everything that was suggested. 😄 

Changed the names of the configurations to be:

* `recommended.json`: aliased by `"extends": ["tslint-microsoft-contrib/recommended"]`
* `latest.json`: aliased by `"extends": ["tslint-microsoft-contrib/latest"]`
* `legacy.json`: aliased by `"extends": ["tslint-microsoft-contrib"]`, with a note that that will eventually be an alias for `recommended.json`. Thoughts?

Rule configs are, for now, still in `configs/` and copied to `dist/build/`, but once the repository changes to in-place (#825), they can be moved to the root directory.

Filed https://github.com/palantir/tslint-react/issues/202 for moving a11y rules into tslint-react. If that doesn't see traction any time soon (it looks like that repository hasn't had much maintenance since June 2018...), we can add the equivalent here. Let's discuss that later on?

src/utils/ExtendedMetadata.ts
```diff
@@ -30,6 +30,6 @@ export type Severity = 'Critical' | 'Important' | 'Moderate' | 'Low';
 export type Level = 'Mandatory' | 'Opportunity for Excellence';
 
 /**
- * Ignored - Use this value to exclude the rule from recommended_ruleset.js and the deployed tslint.json file.
+ * Ignored - Use this value to exclude the rule from recommended.json.
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Yeah that seems reasonable. Changed to the more generic language.

JoshuaKGoldberg:
Whoohoo, thanks for the reviews @IllusionMH & @VincentLanglet! This'll be available in the next release.

Issue #575, opened 2018-10-23
Use the same processes to generate documentation as TSLint core


JoshuaKGoldberg:
Following #573, it's a bit annoying manually crafting a README.md table. If TSLint itself makes the processes it uses to build docs extensible such that we could use it, that'd make it much easier for general community folks to have custom rulesets with good documentation.

As part of this, we should move the detailed explanations from the wiki into rule metadata and samples.

JoshuaKGoldberg:
Huh, [Codelyzer](https://github.com/mgechev/codelyzer) does this: http://codelyzer.com/rules/

...and has a snazzy playground. nice.

JoshuaKGoldberg:
### 💀 _**It's time!**_ 💀

**TSLint is being deprecated**; therefore, it and tslint-microsoft-contrib are no longer accepting pull requests for major new changes or features. See https://github.com/palantir/tslint/issues/4534. 😱

If you'd like to see this change implemented, you have two choices:
* **Recommended:** Check if this is available in [ESLint](https://eslint.org) + [typescript-eslint](https://typescript-eslint.io) ✅
* _Not Recommended: Fork TSLint or tslint-microsoft-contrib locally_ 🤷‍♂️

👋 It was a pleasure open sourcing with you!

_If you believe this message was posted here in error, please comment so we can re-open the issue!_

PR #817, opened 2019-02-09
Strict order of rules in README.md

### LINKED ISSUE #648: Better describe and suggest new rules' version options
Following @IllusionMH's comment in https://github.com/Microsoft/tslint-microsoft-contrib/pull/647#issuecomment-436277440 -  we should have a better template for what users should put in README.md as the rules' starting version.

During the release process, we change them to what's correct, but until then it can be misleading to see incorrect version numbers in the big README.md table.

This should be a small intermediate fix until #575 gets rid of that table.

### LINKED ISSUE #816: Make strict order of rules in README.md
### Feature request

**Is your feature request that we implement a new rule?**
No

**Is your feature request related to a problem? Please describe.**
I've noticed that rules are not sorted alphabetically which might make it harder to find rules.
As was discussed with @JoshuaKGoldberg offline -#575 is preferred fix for documentation, however, enforcing sort order in a meantime is acceptable fix

**Describe the solution you'd like**
Add build task to fail build if rules in `README.md` are out of order.
`create-rule` task should add all markup for rule entry in correct place.

**Describe alternatives you've considered**
Implement #575, but will require more effort.

IllusionMH:
#### PR checklist

-   [X] Addresses an existing issue: fixes #816, fixes #648 
-   [X] New feature
    -   [ ] Includes tests
-   [X] Documentation update

#### Overview of change:
New validation is added to `npm test` that will check if rules in `REAME.md`'s table are sorted alphabetically.
Updated `create-rule` script will also add row for new rule in correct place. `@next` will be used as placeholder for version as was discussed in #648.

I've used regular expression to work with table in `README.md` because HTML markup is pretty simple, RegExp is simple enough, and it helps to avoid adding another dependency for temporary solution (should be removed with  #575).

Also, looks like VS Code no longer sets environment variable in terminal, so files should be opened by hand.

Please check [Releases](https://github.com/Microsoft/tslint-microsoft-contrib/wiki/Releases) that was updated to mention `@next` placeholder.

### DIFF
--- a/build-tasks/common/readme-rules.js
+++ b/build-tasks/common/readme-rules.js
@@ -1,0 +1,22 @@
+const { readFile, writeFile } = require('./files');
+
+// Matches <tr>, <td>, and <code> openning tags with only whitespace surrounding them
+// Capturing group for rule name inside <code> and </code> tags with possible whitespace around it
+const praseRe = /\s+<tr>\s*?<td>\s*?<code>\s*?([0-9a-z-]*?)\s*?<\/code>/g;
+const README_NAME = 'README.md';
+const readmeContent = readFile(README_NAME);
+
+function getAllRules() {
+    const rules = [];
+    let match;
+
+    while ((match = praseRe.exec(readmeContent))) {
+        rules.push(match[1]);
+    }
+
+    return rules;
+}
+
+module.exports = {
+    getAllRules
+};
--- a/build-tasks/common/readme-rules.js
+++ b/build-tasks/common/readme-rules.js
@@ -17,6 +17,34 @@ function getAllRules() {
     return rules;
 }
 
+function addNewRule(name, ruleMarkup) {
+    let match;
+    let insertPosition;
+
+    while ((match = praseRe.exec(readmeContent))) {
+        const existingName = match[1];
+        // Looks for first rule that should follow new rule.
+        // Match position will be used to insert new rule row
+        if (name < existingName) {
+            insertPosition = match.index;
+            break;
+        }
+    }
+
+    // In case when new rule should be added to the end of the table - look for insret position before </tbody>
+    if (insertPosition === undefined) {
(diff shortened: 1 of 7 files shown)


build-tasks/create-rule.js
```diff
@@ -88,13 +90,15 @@ const questions = [
 inquirer.prompt(questions).then(answers => {
     const sourceFileName = createImplementationFile(answers);
     const testFileNames = createTestFiles(answers);
+    const readmePosition = createReadmeEntry(answers);
 
     console.log(`Rule '${answers.name}' created.`);
     console.log(`Source file: ${sourceFileName}`);
-    console.log(`Test files:   ${testFileNames.join(', ')}`);
+    console.log(`Test files:   ${testFileNames.join(' ')}`);
```
IllusionMH: Removed `,` so VS Code terminal will be able to create "link" to the file.

build-tasks/common/readme-rules.js
```diff
@@ -0,0 +1,50 @@
...
+    if (insertPosition === undefined) {
+        insertPosition = readmeContent.match(/\s+<\/tbody>\s+<\/table>/).index;
+    }
+    const contentBeforeNewRule = readmeContent.slice(0, insertPosition);
+    const contentAfterNewRule = readmeContent.slice(insertPosition);
+    const newContent = contentBeforeNewRule + ruleMarkup + contentAfterNewRule;
+    writeFile(README_NAME, newContent);
+
+    // To position cursor on the same line with new rule name should add 3 lines to match lines added with template
+    return `./${README_NAME}:${contentBeforeNewRule.split('\n').length + 3}`;
```
IllusionMH: Without './' VS Code can't create lint to the file in project root.

JoshuaKGoldberg:
This looks great, thanks @IllusionMH! Confirmed Releases.md's mention of `@next` is good.

PR #824, opened 2019-02-17
[react-unused-props-and-state] add support for traversing function components

### LINKED ISSUE #339: react-unused-props-and-state fails with SFC's

shiftkey:
#### PR checklist

-   [x] Addresses an existing issue: fixes #339 
    -   [x] Includes tests
-   [ ] Documentation update

#### Overview of change:

This is an attempt to support finding unused props in function components, based on the usage I've seen in the wild and figuring out how to traverse them properly. Fair warning: this is my first real contribution to anything related using the TypeScript AST in anger, so I want to caveat this with:

Jokes aside, I focused on these two general cases reported in #339.

First, an arrow function that returns a stateless function component. We can check more cases here, like a type literal being used rather than a type reference, because of the context that the presence of the `React.SFC` provides:

```ts
interface Props {
  children: JSX.Element | JSX.Element[];
  className: string;
}

const ButtonGroup: React.SFC<Props> = (props: Props) => {
  return (
    <div className={props.className}>
      {props.children}      
    </div>
  )
};
```

Second, function declarations which have a parameter matching the regex defined in config:

```ts
interface IProps {
	test: string;
	test2: string;
}

export function Test(props: IProps) {
	return <div>{ props.test }</div>;
}
```

I avoided being as flexible as possible because it might incorrectly traverse all functions in a module which have one parameter.

These changes should be additive to how class components are traversed. I didn't want to touch that flow despite the decent suite of tests covering it. Because I wanted these changes to be additive I introduced the `callbackForFunctionComponentNodes` to traverse the interesting parts of these arrow functions and function declarations and find any props usage.

TODO:

 - [x] look for `React.FunctionComponent` as well as `React.SFC` (`SFC` is deprecated in the latest `@types/react` package)

Three things from the reviewers:

 - general feedback about the approach and how it fits with the style of other linting rules
 - being able to build a local NPM package and verify these changes inside a live project using `tslint`
    -  `cp dist/build/*.js ../other-project/node_modules/tslint-microsoft-contrib/` worked for me after b1e907a - is there a better way to do this?
 - ~~the core of this change was about how property expression worked, and I'd like to merge `callbackForFunctionComponentNodes` back into the main `cb` callback to be consistent but haven't quite got the traversal right. Let me know how strongly you'd like for this all to be in `cb`.~~ fixed in 9977484

One thing from the community:

 - I'd love to find additional test cases that aren't covered here which fit with the spirit of the rule. If you can run this PR up locally, add a test case with your snippet and whether it should pass or report errors, and leave a comment in this PR so I can follow up.


src/tests/ReactUnusedPropsAndStateRuleTests.ts
```diff
@@ -614,4 +614,415 @@ describe('reactUnusedPropsAndStateRule', (): void => {
...
+                        name: Utils.absolutePath('file.tsx'),
+                        ruleName: 'react-unused-props-and-state',
+                        startPosition: { character: 19, line: 7 }
+                    }
+                ]);
+            });
+
+            // TODO: this test is a thorny one to resolve, and needs some further thought
+            //       about how to address this with SFCs
+            it.skip('finds missing prop when ObjectBindingExpression used', () => {
```
shiftkey: Putting this back to `WIP` to investigate how I might solve this test case (I could diff the identifiers inside the `ObjectBindingExpression` against `propNames` and if there are values missing, this could immediately report an error).

shiftkey: Should be fixed in b1e907a

IllusionMH:
Haven't chance to look at PR in details yet.
At first glance noticed that there are checks for `React.SFC`, however it is marked as deprecated in favor of `React.FC`. So this rules should check both of this types.

shiftkey:
> So this rules should check both of this types.

@IllusionMH thanks for the tip - I'll put that on my to-do list

src/tests/ReactUnusedPropsAndStateRuleTests.ts
```diff
@@ -614,4 +614,413 @@ describe('reactUnusedPropsAndStateRule', (): void => {
             }
         ]);
     });
+
+    describe('using SFCs', () => {
```
JoshuaKGoldberg: These tests look great!

But - if you're going to add more, you might have an easier time using the new test format (#489). That can be done by adding a new tests folder as in #806 and doesn't necessitate deleting the existing tests in this file. #489 tracks the repository migrating all files that way.

JoshuaKGoldberg:
> general feedback about the approach and how it fits with the style of other linting rules

This seems like a pretty reasonable approach. This rule is an unusually crusty one, which seems inevitable given how vague the definition of what a "component" is. We're very fortunate to have a precedent of props interfaces being named `*Props`...

I think it could later be improved by also considering any arrow function, function declaration, or function expression that takes in a single object and returns anything assignable to `React.ReactNode` or `React.ReactNodeArray`. But that would be good discussion for a followup issue.

> being able to build a local NPM package and verify these changes inside a live project using `tslint`

Yeah the build process for this package is a little wonky. 7.0 will, as a breaking change, remove a lot of older packaging that is no longer relevant, so it'll be possible to switch to in-place file compiles (in other words: no longer need to `cp` files before symlinking them). Until then, you _could_ do something like:

```shell
cd ~/Code/repos/tslint-microsoft-contrib/src
tsc --outDir ~/Code/repos/some-other-project/node_modules/tslint-microsoft-contrib
```

...but you might have to fiddle around with compiler settings to get that to work. 

Personally I just hand-edit what's in `node_modules`. 😅

src/reactUnusedPropsAndStateRule.ts
```diff
@@ -198,6 +230,68 @@ function walk(ctx: Lint.WalkContext<Options>) {
                     stateNames = [];
                 }
             }
+        } else if (tsutils.isTypeReferenceNode(node)) {
+            // This code is interested in any type reference that looks like a
+            // React.SFC<T> because the typical usage here is an arrow function
+            // which accepts a single parameter of type T - this parameter is
+            // the props we need to look for, so we should traverse the body of
+            // this arrow function (if it can be found)
+            if (node.typeName.getText() === 'React.SFC' && node.typeArguments !== undefined && node.typeArguments.length === 1) {
```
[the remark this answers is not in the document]
shiftkey: Should I be stringent here and check that `SFC/StatelessComponent/FC/FunctionalComponent` (without the `React` prefix) were part of the React import? I worry about false positive here (but maybe I'm overthinking things).

src/reactUnusedPropsAndStateRule.ts
```diff
@@ -198,6 +230,68 @@ function walk(ctx: Lint.WalkContext<Options>) {
...
+
+                if (typeArgument && tsutils.isTypeReferenceNode(typeArgument)) {
+                    const typeName = typeArgument.typeName.getText();
+                    if (ctx.options.propsInterfaceRegex.test(typeName)) {
+                        propsAlias = firstParameter.name.getText();
+                        functionComponents.push(node.body);
+                    }
+                }
+            }
+        } else if (tsutils.isReturnStatement(node) && node.expression) {
```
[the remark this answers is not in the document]
shiftkey: I left this around because I wasn't sure if `node.expression` is considered a child. I'll test it out and report back.

src/tests/ReactUnusedPropsAndStateRuleTests.ts
shiftkey: After playing around with this I ended up just porting the old tests in #826, and I'll rebase this PR on top of that so all of the tests in this PR are also written using the inbuilt `tslint` test framework.

src/reactUnusedPropsAndStateRule.ts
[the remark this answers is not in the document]
shiftkey: Cleaned up without problems in 95ca41b

src/reactUnusedPropsAndStateRule.ts
```diff
@@ -198,6 +230,68 @@ function walk(ctx: Lint.WalkContext<Options>) {
...
+
+                            for (const propName of foundPropNames) {
+                                propNames = Utils.remove(propNames, propName);
+                            }
+                        }
+                        arrowFunctions.push(arrowFunction);
+                    }
+                }
+            }
+        } else if (tsutils.isFunctionDeclaration(node) && node.body) {
```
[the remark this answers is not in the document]
shiftkey: Done in 33e0737241b4a2202d8760c43ae37f159086ecd4 and 44e064e0febc1299a1148a54ef0bb6b1624b58e3

shiftkey:
I think this is ready for another round of review

src/reactUnusedPropsAndStateRule.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: No, let's not bother checking that. Some projects define those in their own typings so we don't want to exclude them.

src/reactUnusedPropsAndStateRule.ts
```diff
@@ -104,6 +136,109 @@ function walk(ctx: Lint.WalkContext<Options>) {
...
+    }
+
+    function lookForReactSpecificArrowFunction(node: ts.TypeReferenceNode): void {
+        const nodeTypeText = node.typeName.getText();
+
+        const isFunctionComponent =
+            nodeTypeText === 'React.SFC' ||
+            nodeTypeText === 'React.FC' ||
+            nodeTypeText === 'React.StatelessComponent' ||
+            nodeTypeText === 'React.FunctionComponent';
```
JoshuaKGoldberg: Could you add some a test case for each of these? They probably don't have to be as extensive as the existing ones, but there should be just enough to verify that these strings are in fact checked.

JoshuaKGoldberg: Per the other thread, are you averse to also checking `'SFC'`, `'FC'`, etc.? It's not a blocker, but IMO would be nice.

shiftkey: Not a problem at all - I'll add a check for the specific names as well...

shiftkey: I ended up getting overwhelmed at how the tests were organized while tackling this and ended up breaking them down into more specific areas (this also helped show which areas were better covered than others:



The new tests were named (for both the old `SFC`/`StatelessComponent` and the new `FC`/`FunctionComponent` APIs: 

 - `reports-for-explicit-import-shorthand.tsx.lint`
 - `reports-for-explicit-import.tsx.lint`

src/reactUnusedPropsAndStateRule.ts
```diff
@@ -208,6 +355,8 @@ function walk(ctx: Lint.WalkContext<Options>) {
     // If there are Props or State interfaces, then scan the classes now.
```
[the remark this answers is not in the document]
shiftkey: Updated in 87be7d7

shiftkey:
@IllusionMH lets open a fresh issue or two to capture these other gaps - I'd rather not hold up getting this into an update for the new cases it currently covers.

IllusionMH:
@shiftkey OK, I'll create separate issues for mentioned cases.

Thanks you for this contribution!

PR #829, opened 2019-02-20
Add case sensitivity to import-name rule

### LINKED ISSUE #830: Add case sensitivity to 'import-name' rule
### Rule improvement suggestion

This rule improvement is specific for my development style, but it would not affect general style.

#### Why
Let's say, there is a rule for file naming using `dots.case` and a rule to use `export default` only for `constructor`s. So importing variable should be named in `PascalCase`. Current `import-name` rule would not allow this.

#### Suggestion
I suggest adding an option to set the preferred string case with `camelCase` by default.

#### Use case
```ts
// ./src/master.class.ts
export default class Master {}
```
```ts
// ./src/another.file.ts
import Master from './master.class';
class Slave extends Master {} 
```

djomaa:
#### PR checklist

-   [x] Addresses an existing issue: fixes #830
    -   [x] Includes tests
-   [ ] Documentation update

#### Overview of change:
`import-name` update. It gives a possibility to set string case format (`camelCase` | `pascalCase`).

###### Without update
```
// failure
import AbstractController from './controllers/abstract.controller';
```
```
// success
import abstractController from './controllers/abstract.controller';
```
###### With update
* camelCase
  * `"import-name": [true, {}, {}, { case: "camelCase" }],`
  * it's a default value
```
// failure
import AbstractController from './controllers/abstract.controller';
```
```
// success
import abstractController from './controllers/abstract.controller';
```
* PascalCase
  * `"import-name": [true, {}, {}, { case: "PascalCase" }],`
```
// failure
import abstractController from './controllers/abstract.controller';
```
```
// success
import AbstractController from './controllers/abstract.controller';
```

### DIFF
--- a/src/importNameRule.ts
+++ b/src/importNameRule.ts
@@ -83,11 +83,11 @@ export class Rule extends Lint.Rules.AbstractRule {
 
     private extractConfig(opt: unknown): Config {
         const result: Config = { ignoreExternalModule: true, case: StringCase.camel };
-        const configKeyLlist: ConfigKey[] = ['ignoreExternalModule', 'case'];
+        const configKeyList: ConfigKey[] = ['ignoreExternalModule', 'case'];
         if (isObject(opt)) {
             return Object.keys(opt).reduce(
                 (accum: Config, key: string) => {
-                    if (configKeyLlist.filter((configKey: string) => configKey === key).length >= 1) {
+                    if (configKeyList.filter((configKey: string) => configKey === key).length >= 1) {
                         accum[<ConfigKey>key] = <boolean | StringCase>opt[key];
                         return accum;
                     }
@@ -201,7 +201,7 @@ function walk(ctx: Lint.WalkContext<Option>) {
             if (makeCamelCase(expectedImportedName) === importedName || makePascalCase(expectedImportedName) === importedName) {
                 return true;
             }
-        } else if (transformName(expectedImportedName) === importedName) {
+        } else if (transformName(expectedImportedName).includes(importedName)) {
             return true;
         }
 
@@ -226,11 +226,11 @@ function walk(ctx: Lint.WalkContext<Option>) {
     function transformName(input: string) {
         switch (option.config.case) {
             case StringCase.camel:
-                return makeCamelCase(input);
+                return [makeCamelCase(input)];
             case StringCase.pascal:
-                return makePascalCase(input);
+                return [makePascalCase(input)];
             case StringCase.any:
-                return `${makeCamelCase(input)} or ${makePascalCase(input)}`;
+                return [makeCamelCase(input), makePascalCase(input)];
             default:
                 throw new Error(`Unknown case for import-name rule: "${option.config.case}"`);
         }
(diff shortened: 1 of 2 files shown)


IllusionMH:
@djomaa thank you for your contribution! Usually any PRs are accepted if there is related issue with clear problem description and agrred solution. Otherwise it might be hard to review when intent isn't clear.
If your changes are for existing issue - please add issu to PR checklist. If you propose something that wasn't previously discussed - please create new issue with problem description and proposal.

P.S. It's a bit unclear why same code in "Without update" section is mentioned as failure and success.

djomaa:
@IllusionMH ok, sorry, i'll create an issue.

> P.S. It's a bit unclear why same code in "Without update" section is mentioned as failure and success.  

My bad, that's corrected. Thank you.

IllusionMH:
@JoshuaKGoldberg what do you think about providing third option `any` as described  in https://github.com/Microsoft/tslint-microsoft-contrib/issues/830#issuecomment-467524087

JoshuaKGoldberg:
@IllusionMH oh, right - I'm in favor, though the name might want to include a hint that it's about casing (`any-case`)?

djomaa:
@IllusionMH @JoshuaKGoldberg yeah, i like `any-case`. I can add changes in the next 3-4 days.

IllusionMH:
@djomaa this would be great. Thanks!

JoshuaKGoldberg:
Ping @djomaa - do you have time to make the changes? We can if not :)

djomaa:
@JoshuaKGoldberg sorry, I have no time. Would be great if u finished this.

djomaa:
`any-case` is ready

src/importNameRule.ts
```diff
@@ -205,6 +223,19 @@ function walk(ctx: Lint.WalkContext<Option>) {
...
     }
 
+    function transformName(input: string) {
+        switch (option.config.case) {
+            case StringCase.camel:
+                return makeCamelCase(input);
+            case StringCase.pascal:
+                return makePascalCase(input);
+            case StringCase.any:
+                return `${makeCamelCase(input)} or ${makePascalCase(input)}`;
```
IllusionMH: Output of `transformName` will be used as replacement in fix and this string will produce incorrect TS code.

src/tests/ImportNameRuleTests.ts
```diff
@@ -271,4 +241,119 @@ describe('importNameRule', (): void => {
...
+            it('should pass on PascalCase and camelCase', () => {
+                const script: string = `
+                    import App from 'app';
+                    import app from 'app';
+                    import App from 'x/y/z/App';
+                    import GraphqlTag from 'x/y/z/graphql-tag';
+                    import graphqlTag from 'x/y/z/graphql-tag';
+                `;
+                const options = [true, {}, {}, { case: 'any-case' }];
+                TestHelper.assertViolationsWithOptions(ruleName, options, script, []);
```
IllusionMH: I think it will be nice to have example with error for `any-case` options.

P.S. Error case would also show problem with fix in this rule.

src/importNameRule.ts
```diff
@@ -81,28 +82,36 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 
     private extractConfig(opt: unknown): Config {
-        const result: Config = { ignoreExternalModule: true };
-        const configKeyLlist: ConfigKey[] = ['ignoreExternalModule'];
+        const result: Config = { ignoreExternalModule: true, case: StringCase.camel };
+        const configKeyLlist: ConfigKey[] = ['ignoreExternalModule', 'case'];
```
JoshuaKGoldberg: Nitpick: `configKeyList` :)

src/importNameRule.ts
```diff
@@ -183,7 +197,11 @@ function walk(ctx: Lint.WalkContext<Option>) {
         moduleName: string,
         node: ts.ImportEqualsDeclaration | ts.ImportDeclaration
     ): boolean {
-        if (expectedImportedName === importedName) {
+        if (option.config.case === StringCase.any) {
```
IllusionMH: This is las one from my side.
With `transformName` + `indexOf` (yeah, I forgot about Node 6) there is no need for special handling of `StringCase.any`.
`transformName(expectedImportedName).indexOf(importedName) > -1` should catch all cases now.

IllusionMH:
@djomaa looks great now! 👍 Thank you for your contribution and following along with review!

IllusionMH:
Since there was only one change requested from @JoshuaKGoldberg and it was fixed, I'll merge this one.

PR #832, opened 2019-02-23
convert no-jquery-raw-elements rule to use a walk function

drexler:
#### PR checklist

#### Overview of change:
Converts no-jquery-raw-elements rule to use a walk function


JoshuaKGoldberg:
Whoops, merging the PR must not have gone through...

PR #847, opened 2019-03-13
Update w3 reference in the react-a11y-role rule

### LINKED ISSUE #844: Update w3 reference in the react-a11y-role rule
Currently, the `react-a11y-role` rule points to an [old URL](https://www.w3.org/TR/wai-aria/roles#role_definitions) that doesn't exist anymore.

I never saw the previous URL, but I believe the content was the same contained in [this URL](https://www.w3.org/TR/wai-aria-1.1/#role_definitions).

However, we also have [this new URL](https://www.w3.org/WAI/PF/aria/roles#roles_categorization) that seems to contain a more complete explanation about each one of the possible `role` values.

Should we update the URL in the `react-a11y-role` rule? If so, to which URL?

I'd be glad to open a PR with this change, if possible.

arthurdenner:
#### PR checklist

-   [x] Addresses an existing issue: closes #844
    -   [x] Includes tests
-   [x] Documentation update

#### Overview of change:
Updates the reference to role definitions in the `react-a11y-role` rule.

No.


JoshuaKGoldberg:
Perfect. Thanks!

PR #849, opened 2019-03-14
Mouse events have key events rule

### LINKED ISSUE #279: new a11y rule: mouse-events-have-key-events
mouse-events-have-key-events

Enforce that onMouseOver/onMouseOut are accompanied by onFocus/onBlur for keyboard-only users.
    • http://oaa-accessibility.org/wcag20/rule/59/
    • http://oaa-accessibility.org/wcag20/rule/60/

onmousedown, onmouseup and onmousemove event handlers should have keyboard equivalents.
    • http://oaa-accessibility.org/wcag20/rule/63/

Check that elements with mouse event handlers also have key event handlers
    • http://oaa-accessibility.org/wcag20/rule/95/

Check that elements with ondblclick handles also have key event handlers
    • http://www.w3.org/TR/WCAG20-TECHS/SCR20
    • From HTML Code Sniffer (https://squizlabs.github.io/HTML_CodeSniffer/Standards/WCAG2/)

lizzzp1:
#### PR checklist

-   [x] Addresses an existing issue: fixes #279 
    -   [x] Includes tests
-   [ ] Documentation update

#### Overview of change:
https://github.com/Microsoft/tslint-microsoft-contrib/issues/279
Adds rule for` Enforce that onMouseOver/onMouseOut are accompanied by onFocus/onBlur for keyboard-only users.`
Planning on adding the next few, but starting here to make sure this is heading in the right direction.


README.md
```diff
@@ -938,6 +938,15 @@ We recommend you specify exact versions of lint libraries, including `tslint-mic
...
       <td>2.0.11</td>
     </tr>
+    <tr>
+      <td>
+        <code>react-a11y-mouse-event-has-key-event</code>
+      </td>
+      <td>
+        For accessibility of your website, elements with mouseOver/mouseOut should be accompanied by onFocus/onBlur keyboard events.
+      </td>
+      <td>6.1.0</td>
```
lizzzp1: @JoshuaKGoldberg what is the next version. I think I need to catch this branch up to master.

lizzzp1: Also how do you generate the `tslint-warnings.csv` from this? I know I did this before 🤔

IllusionMH: @lizzzp1 new rules should have `@next` in version column. It will be replaced during development.

csv file should be created from rule class metadata.

lizzzp1: @IllusionMH ah, got it. Thanks!

src/reactA11yMouseEventHasKeyEventRule.ts
```diff
@@ -0,0 +1,109 @@
...
+const MOUSE_EVENTS: {
+    onMouseOver: {
+        value: 'onmouseover';
+        jsxValue: 'onMouseOver';
+    };
+    onMouseOut: {
+        value: 'onmouseout';
+        jsxValue: 'onMouseOut';
+    };
+} = {
```
lizzzp1: Seems like these literals don't get inferred by the compiler. Wondering if there's a less verbose way to declare these.

JoshuaKGoldberg: How about creating an interface for the two objects containing `value` and `jsxValue`?

Alternately, perhaps with the new `as const` (`const` assertions)? https://devblogs.microsoft.com/typescript/announcing-typescript-3-4-rc/ Great timing if so!

Suggestion: upgrade tslint-microsoft-contrib to the 3.4 RC, and if that solves your use case, we can put tslint-microsoft-contrib onto that version before this PR goes in? /cc @IllusionMH

IllusionMH: Interface will have type `string` instead of literals.

Until 3.4 arrives explicit type assertions to the same literal might be a better way
```ts
const MOUSE_EVENTS = {
  onMouseOver: {
    value: 'onmouseover' as 'onmouseover',
    jsxValue: 'onMouseOver' as 'onMouseOver'
  },
  onMouseOut: {
    value: 'onmouseout' as 'onmouseout',
    jsxValue: 'onMouseOut' as 'onMouseOut'
  }
};
```

Frankly speaking, I don't know if it safe to jump 3 versions ahead of `tslint` package with shared dependency on `tsutils` and yet unknown Prettier and TSLint support for mentioned `const` type assertions.
If there shouldn't be any problems (shouldn't be a big problem with manual AST traversing) - I have no objections to upgrade TS.

tests/react-a11y-mouse-event-has-key-event/test.tsx.lint
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react'
```
JoshuaKGoldberg: Missing test cases, none of which should have errors _(right?)_:

* `<div />`
* `<div {...props} />`
* `<div onMouseOver={() => {}} {...props} />`
* `<div onMouseOut={() => {}} {...props} />`

src/reactA11yMouseEventHasKeyEventRule.ts
```diff
@@ -0,0 +1,109 @@
...
+            mouseEvent: typeof MOUSE_EVENTS.onMouseOver | typeof MOUSE_EVENTS.onMouseOut,
+            focusEvent: typeof FOCUS_EVENTS.onBlur | typeof FOCUS_EVENTS.onFocus
+        ): void {
+            const attributes: AttributeType = getJsxAttributesFromJsxElement(node);
+
+            if (attributes === undefined) {
+                return;
+            }
+
+            const attributeKeys = Object.keys(attributes);
```
JoshuaKGoldberg: Nit: could use a `Set` so the later lookups are O(1), in case of elements with many attributes. 
```suggestion
            const attributeKeys = new Set(Object.keys(attributes));
```

lizzzp1: Ooo I like this suggested changes feature.
Good call.

tests/react-a11y-mouse-event-has-key-event/test.tsx.lint
lizzzp1: So the last two do complain, but we'd have to do the type checking in order to extract out those values and check the spread props. Right?

JoshuaKGoldberg: Sure, if you want! Two options here:

* Don't do that, and file a followup issue tracking making this an (optionally) type checked rule
* Do that as part of this PR

If&whenever that goes in, the rule should be an `OptionallyTypedRule`, like `unneccessary-bind` is from TSLint core. That way folks can still use the rule if running without a `tsconfig.json` project. 

Does that answer your question?

lizzzp1: > followup issue tracking making this an (optionally) type checked rule

that works

src/reactA11yMouseEventHasKeyEventRule.ts
```diff
@@ -0,0 +1,109 @@
...
+    };
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithFunction(sourceFile, walk);
+    }
+}
+
+function walk(ctx: Lint.WalkContext<void>) {
+    function cb(node: ts.Node): void {
+        function checkMouseEventForFocus(
```
IllusionMH: Function declaration inside `cb` that will be called for each node in file looks not optimal.

Better to move it to `walk` and pass `node` as an argument.

lizzzp1: 👍

src/reactA11yMouseEventHasKeyEventRule.ts
lizzzp1: This complains about `as-cast` 🤔

JoshuaKGoldberg: Classic TSLint...

tests/react-a11y-mouse-event-has-key-event/test.tsx.lint
```diff
@@ -0,0 +1,15 @@
...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~             [onMouseOver must be accompanied by onFocus.]
+const elementTwo = (<div onMouseOut={() => {}}>click</div>)
+                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~             [onMouseOut must be accompanied by onBlur.]
+const elementSelfClosing = (<div onMouseOut={() => {}}/>)
+                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~  [onMouseOut must be accompanied by onBlur.]
+
+<div onMouseOver={() => {}} {...props} />
+<div onMouseOut={() => {}} {...props} />
+<div />
+<div {...props} />
```
JoshuaKGoldberg: These four cases, where the line starts with `<`, are invalid TypeScript. You'll need to store these in a variable as with the other cases.

lizzzp1: Oh oops.
Ok, so
```
const foo = <div onMouseOver={() => {}} {...props} />
const bar = <div onMouseOut={() => {}} {...props} />
``` 
does complain, which makes sense. So do you just ignore elements with spread attributes?

JoshuaKGoldberg: I think so, yeah. Without a program's type checker there's no way to know if they're valid.

tests/react-a11y-mouse-event-has-key-event/test.tsx.lint
```diff
@@ -7,9 +7,11 @@ const elementTwo = (<div onMouseOut={() => {}}>click</div>)
...
                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~  [onMouseOut must be accompanied by onBlur.]
 
-<div onMouseOver={() => {}} {...props} />
-<div onMouseOut={() => {}} {...props} />
-<div />
-<div {...props} />
+const Foo = (<div onMouseOver={() => {}} {...props} />)
+            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [onMouseOver must be accompanied by onFocus.]
+const Bar = (<div onMouseOut={() => {}} {...props} />)
+            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [onMouseOut must be accompanied by onBlur.]
```
lizzzp1: Complains for now.

JoshuaKGoldberg: @lizzzp1 could you change it to not complain if there's a `...`, actually? It would be annoying to have a `...props` that includes the correct corresponding `onBlue`/`onFocus` and still have this complain.

lizzzp1: I thought that required the type checker that was a follow up? I'm really confused now.

JoshuaKGoldberg: The type checker is just required to know whether the `...props` contains the corresponding `onBlur`/`onFocus`. Without the type checker, we have know way of knowing whether a JSX element that contains `...props` violates the rule or not. Until the type checker logic is added, that leaves us with two options:

* Ignore `...props` _(pretend they aren't therE): the rule would give complaints in some places where it shouldn't
* Assume `...props` contains the right attributes: the rule wouldn't give complaints in some places where it should

In order to not irritate folks with false complaints, IMO we should err on the side of caution with the first option.

Does that seem reasonable?

lizzzp1: Ok, if props are ignored then it would complain like the above wouldn't it? Wouldn't the 2nd option be better so it doesn't complain if it encounters spread props?

JoshuaKGoldberg: Oh agh sorry I _meant_ to say the second option! You're absolutely correct 😅

PR #858, opened 2019-05-01
ignore input type="file" from react-a11y-input-elements 

### LINKED ISSUE #841: Rule react-a11y-input-elements incorrectly requiring value for file input
### Bug Report

-   **`tslint-microsoft-contrib` version**: 6.1.0 
-   **TSLint version**: 5.13.1
-   **TypeScript version**: 3.3.3333
-   **Running TSLint via**: CLI

```tsx
import * as React from "react";

export const X: React.ReactNode = (
  <input
    type="file"
    accept="image/*"
    placeholder="this is file input"
  />
);
```

```json
defaultSeverity: error
extends:
  - tslint-microsoft-contrib
  - tslint-config-prettier
jsRules:
rules:
```

#### Actual behavior
`ERROR: (react-a11y-input-elements) /....../X.tsx[4, 5]: Input elements must include default, place-holding characters if empty`

#### Expected behavior
no error

or

What value should  I set to value attribute ?

umezo:
#### PR checklist

-   [x] Addresses an existing issue: fixes #841
-   [x] bugfix
    -   [x] Includes tests
-   [ ] Documentation update

#### Overview of change:
Ignore any `<input type="file">` from react-a11y-input-elements because both `placeholder` and `value` attributes are ignored by browsers or screen readers. 

Should I update any document ?

### DIFF
--- a/src/reactA11yInputElementsRule.ts
+++ b/src/reactA11yInputElementsRule.ts
@@ -51,12 +51,14 @@ function isTypeMatchedTo(
     attributes: { [propName: string]: ts.JsxAttribute },
     condition: (attributeText: string) => boolean
 ): boolean {
+    if (attributes.type === undefined) {
+        return false;
+    }
     for (const attribute of node.attributes.properties) {
         if (tsutils.isJsxAttribute(attribute)) {
-            const isInputAttributeType = attributes.type;
             if (attribute.initializer !== undefined && tsutils.isStringLiteral(attribute.initializer)) {
                 const attributeText = attribute.initializer.text;
-                if (isInputAttributeType !== undefined && condition(attributeText)) {
+                if (condition(attributeText)) {
                     return true;
                 }
             }
--- a/src/reactA11yInputElementsRule.ts
+++ b/src/reactA11yInputElementsRule.ts
@@ -46,13 +46,17 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 }
 
-function isExcludedInputType(node: ts.JsxSelfClosingElement, attributes: { [propName: string]: ts.JsxAttribute }): boolean {
+function isTypeMatchedTo(
+    node: ts.JsxSelfClosingElement,
+    attributes: { [propName: string]: ts.JsxAttribute },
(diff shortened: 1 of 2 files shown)


src/reactA11yInputElementsRule.ts
```diff
@@ -61,6 +61,21 @@ function isExcludedInputType(node: ts.JsxSelfClosingElement, attributes: { [prop
     return false;
 }
 
+function isInputTypeFile(node: ts.JsxSelfClosingElement, attributes: { [propName: string]: ts.JsxAttribute }): boolean {
```
JoshuaKGoldberg: This method is almost the same as `isExcludedInputType`; the only difference is in the innermost `if` statement. You should be able to deduplicate the code a bit by passing in a `condition: (attributeTex: string) => boolean` to the method, or something similar. 

_(if you can think of a more elegant way, that'd be great!)_

src/reactA11yInputElementsRule.ts
```diff
@@ -61,6 +61,21 @@ function isExcludedInputType(node: ts.JsxSelfClosingElement, attributes: { [prop
...
+    for (const attribute of node.attributes.properties) {
+        if (tsutils.isJsxAttribute(attribute)) {
+            const isInputAttributeType = attributes.type;
```
JoshuaKGoldberg: > `isInputAttributeType`

This is a bit of an existing concern, but the inner two `if` statements are not necessary if `isInputAttributeType` is falsy. Could you change these functions to return before the `for` loop if that's the case? Seems like a lot of wasted effort here.

src/reactA11yInputElementsRule.ts
umezo: I didn't come up with anything 😁 but agree with your idea.

src/reactA11yInputElementsRule.ts
umezo: Oh, I couldn't find it. These codes are a little complex to read.

umezo:
@JoshuaKGoldberg Thank you for review. I applied your comments.

PR #862, opened 2019-05-12
Improve prefer-array-literal rule

### LINKED ISSUE #822: prefer-array-literal: how can one specify the size of the array?
how can one specify the size of the array?

const foo = new Array(10)
?

_Originally posted by @ogiexela in https://github.com/Microsoft/tslint-microsoft-contrib/issues/110#issuecomment-463734744_

### LINKED ISSUE #856: False positive for prefer-array-literal when using `Array` from other namespace
### Bug Report

-   **`tslint-microsoft-contrib` version**: 6.1.0-beta2
-   **TSLint version**: 5.12.1
-   **TypeScript version**: 3.3.3333
-   **Running TSLint via**:  VSCode

```tsx
import { Types } from 'mongoose';

export const foo: Types.Array<number> = new Types.Array();
```

```json
  "extends": ["tslint-microsoft-contrib"],
  "rulesDirectory": ["node_modules/tslint-microsoft-contrib"],
  "rules": {
    "trailing-comma": false,
    "no-relative-imports": false,
    "no-implicit-dependencies": [true, ["dev", "graphql-tools"]],
    "newline-per-chained-call": false,
    "no-console": false,
    "quotemark": false,
    "export-name": [true, { "allow": ["resolvers"] }],
    "no-backbone-get-set-outside-model": false,
    "no-unsafe-any": false,
    "strict-boolean-expressions": false,
    "typedef": [
      true,
      "call-signature",
      "arrow-call-signature",
      "parameter",
      // "arrow-parameter",
      "property-declaration",
      "variable-declaration",
      "member-variable-declaration",
      "object-destructuring",
      "array-destructuring"
    ],
    "match-default-export-name": false
  }
```

#### Actual behavior

Got `Replace Array constructor with an array literal: new Types.Array() (prefer-array-literal)tslint(prefer-array-literal)`

#### Expected behavior

Got nothing

IllusionMH:
#### PR checklist

-   [X] Addresses an existing issue: fixes #856, fixes #822
-   [X] New feature, bugfix, or enhancement
    -   [X] Includes tests
-   [X] Documentation update

#### Overview of change:
1. Updated tslint test for rule
1. It is forbidden to call `Array` only like `Array`, `window.Array`, `global.Array`, and `globalThis.Array`. Other namespaces are allowed.
1. Allow calls with single number parameter

I haven't included `self` from Web Wrokers to list of restricted parents for property access expression because of `var self = this` pattern. Should I include `self` too?

Any suggesting what should trigger error for single parameter if type checker is not available?
I think that allowing everything without type checker (like ESLint does) and expect anything assignable to number if checker available will be good enough.

Code is ready for review, ~but I still need to add options description to rule meta and update README.md with new option.~


src/preferArrayLiteralRule.ts
```diff
@@ -26,15 +38,26 @@ export class Rule extends Lint.Rules.AbstractRule {
     };
 
     public static GENERICS_FAILURE_STRING: string = 'Replace generic-typed Array with array literal: ';
-    public static CONSTRUCTOR_FAILURE_STRING: string = 'Replace Array constructor with an array literal: ';
-    public static FUNCTION_FAILURE_STRING: string = 'Replace Array function with an array literal: ';
+    public static getReplaceFailureString = (type: InvocationType, nodeText: string) =>
+        `Replace Array ${type} with an array literal: ${nodeText}`;
+    public static getSingleParamFailureString = (type: InvocationType, nodeText: string) =>
+        `To create an array of given length you should use non-negative integer. Otherwise replace Array ${type} with an array literal: ${nodeText}`;
```
IllusionMH: Would be glad to hear suggestions on better message.

src/preferArrayLiteralRule.ts
```diff
@@ -47,18 +70,40 @@ export class Rule extends Lint.Rules.AbstractRule {
 
         ruleOptions.forEach((opt: unknown) => {
             if (isObject(opt)) {
-                value = opt['allow-type-parameters'] === true;
+                allowSingleArgument = opt['allow-single-argument'] === true;
```
IllusionMH: Is this option name OK?

JoshuaKGoldberg: Hmm, if the purpose of the argument is to allow sized creation, how about `allow-size-argument`? `single` might be interpreted as implying a more broad set of allowed behavior.

src/preferArrayLiteralRule.ts
```diff
@@ -6,16 +6,44 @@ import { AstUtils } from './utils/AstUtils';
 import { ExtendedMetadata } from './utils/ExtendedMetadata';
 import { isObject } from './utils/TypeGuard';
 
+// undefined for case when function/constructor is called directly without namespace
+const RESTRICTED_NAMESPACES = [undefined, 'window', 'global', 'globalThis'];
```
JoshuaKGoldberg: > `self = this`

IMO `"self"` should be included here. There's a separate lint rule to exclude the `var self = this;` pattern, which is a holdover from old now-unnecessary (and hopefully now uncommon) styles.

JoshuaKGoldberg: Nit: just for semantic correctness, how about a `Set`?

src/preferArrayLiteralRule.ts
JoshuaKGoldberg: I can't think of anything significantly better :smile: other than adding an `a` before `given`.

```
        `To create an array of a given length you should use a non-negative integer. Otherwise replace Array ${type} with an array literal: ${nodeText}`;
```

Also, do these need to print `nodeText`? Seems redundant.

src/preferArrayLiteralRule.ts
IllusionMH: OK. Wan't aware about rule.

IllusionMH: Yeah, sure. Always remember that Node 6 doesn't support `includes` but forgot about `Set`. Will update implementation

src/preferArrayLiteralRule.ts
IllusionMH: I'm really not a fan of adding `nodeText` to an error messages (not only in this rule) because node will be fully underlined in editor when file opened.
However I've added node text because other messages have one. Should I update all messages to drop node text or just this one?

src/preferArrayLiteralRule.ts
IllusionMH: Yep, this one makes it more clear.

src/preferArrayLiteralRule.ts
IllusionMH: Removed `nodeText` from new error message.
There is still question if I should remove it from other messages.

src/preferArrayLiteralRule.ts
```diff
@@ -47,18 +86,40 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+            const callArguments = node.arguments;
+            if (!allowSizeArgument || !callArguments || callArguments.length !== 1) {
+                const failureString = Rule.getReplaceFailureString(type, node.getText());
+                ctx.addFailureAt(node.getStart(), node.getWidth(), failureString);
+            } else {
+                // When typechecker is not available - allow any call with single expression
+                if (checker) {
+                    const argument = callArguments[0];
+                    const argumentType = checker.getTypeAtLocation(argument);
+                    if (!tsutils.isTypeAssignableToNumber(checker, argumentType) || argument.kind === ts.SyntaxKind.SpreadElement) {
```
IllusionMH: It was unexpected for me that when I used spread of number array type there was no error, however there was an error if I used array of string elements. Added additional check where type comparison wasn't enough.

@JoshuaKGoldberg do you think that check can be left here or should it be moved to previous condition (`if (!allowSizeArgument || !callArguments...`)? If moved - spread element in argument position will be always restricted which is also seems logical.

src/preferArrayLiteralRule.ts
JoshuaKGoldberg: Ah, talked offline, I missed that bit in the review here - burn them all! 🔥

src/preferArrayLiteralRule.ts
JoshuaKGoldberg: That seems reasonable. A little unexpected (perhaps a comment is warranted?)

IllusionMH: I'll add comment.
During offline discussion decided to leave checks in current state.

src/preferArrayLiteralRule.ts
```diff
@@ -115,6 +115,9 @@ function walk(ctx: Lint.WalkContext<Options>, checker: ts.TypeChecker | undefine
                         // See https://github.com/microsoft/tslint-microsoft-contrib/issues/859
                         const argument = callArguments[0];
                         const argumentType = checker.getTypeAtLocation(argument);
+                        // Looks like TS returns type or array elements when SpreadElement is passed to getTypeAtLocation.
+                        // Additional check for SpreadElement is required to catch case when array has type number[] and its element will pass assignability check.
+                        // SpreadElement shouldn't be allowed because result of `Array(...arr)` call will dependepend on array lenght and might produce unexpected results.
```
IllusionMH: Hope this comment makes intention more clear.

JoshuaKGoldberg: Amusing and yes :)

IllusionMH:
Thank you for feedback and review!

PR #501, opened 2018-09-18
Add options along with tests for the function-name rule

ckhatri:
Related to issue #498.

If y'all have any suggestions, let me know! 

Edit: Also auto fixed all the " to ' for the linter in FunctionNameRuleTests.ts

PR #877, opened 2019-06-10
fix: In function-name, actually run static regex for static private when option indicates that it should

### LINKED ISSUE #661: validate-private-statics-as-either skips some names and needs tests
In working on migrating `function-name` tests as part of #489, I noticed that the test cases for options such as `validate-private-statics-as-either` don't test that the rule gives complaints when those options are enabled. Seems like a gap in test coverage.

Anybody is welcome to tackle this! 😊

arinwt:
Fixing function-name rule functionality added by PR #501.  These options should be reworked in the future, but the intent of this change is that the existing intended functionality is met.

#### PR checklist

-   [x] Addresses an existing issue: fixes #661
    -   [x] Includes tests
-   [x] Documentation update

#### Overview of change:

The function-name rule supports options for handling private/protected static methods.  The default option is "validate-private-statics-as-private", which means that private/protected static methods will only validate against the private-method regex.

When changing to the either the "validate-private-statics-as-either" or "validate-private-statics-as-static" option, it would not perform and validation on private/protected static methods.

After the change, "validate-private-statics-as-static" will only validate against the static-method regex, and "validate-private-statics-as-either" will validate against both, but only add a failure if both fail to match.

### DIFF
--- a/src/functionNameRule.ts
+++ b/src/functionNameRule.ts
@@ -128,25 +128,50 @@ function walk(ctx: Lint.WalkContext<Options>) {
     function cb(node: ts.Node): void {
         if (tsutils.isMethodDeclaration(node)) {
             const name: string = node.name.getText();
-            if (AstUtils.hasComputedName(node)) {
+
+            const isComputed = AstUtils.hasComputedName(node);
+            const isPrivate = AstUtils.isPrivate(node);
+            const isProtected = AstUtils.isProtected(node);
+            const isStatic = AstUtils.isStatic(node);
+
+            const staticOnly = isStatic && validateStatics === VALIDATE_PRIVATE_STATICS_AS_STATIC;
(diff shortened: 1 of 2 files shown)


src/functionNameRule.ts
```diff
@@ -128,25 +128,50 @@ function walk(ctx: Lint.WalkContext<Options>) {
     function cb(node: ts.Node): void {
         if (tsutils.isMethodDeclaration(node)) {
```
JoshuaKGoldberg: These `if` blocks are getting pretty huge, and have a lot of variables declared in them. This rule's ever-increasing complexity is likely a partial cause of the option being overlooked before - it's really hard to keep track of the logic here!

Let's split these blocks into separate functions.

arinwt: Thanks for the feedback.  I was trying to do minimal changes, maintaining the existing logical flow and structure until a cleaner rework of these options came about.

Anyway, I have updated the code to break the validations out into separate functions and have different handlers for each configuration.  I am open to changing it further with more specific instructions.

Issue #881, opened 2019-07-07
import-name rule renaming map is not working


alexeychikk:
### Bug Report

-   **`tslint-microsoft-contrib` version**: 6.2.0
-   **TSLint version**: 5.18.0
-   **TypeScript version**: 3.5.2
-   **Running TSLint via**: VSCode

#### TypeScript code being linted

```tsx
import React from 'react';
```

with `tslint.json` configuration:

```json
{
  "extends": "tslint-microsoft-contrib",
  "rules": {
    "import-name": [
      true,
      { "react": "React" }
    ]
  }
}

```

#### Actual behavior
`import React` is being replaced with `import react` on save.
 
#### Expected behavior
Nothing should happen because I defined `{ "react": "React" }` in my tslint.json.

dosentmatter:
Is your config extending anything else? I had the same issue but I found that the tslint-config-airbnb@5.11.1 I was using depended on tslint-microsoft-contrib@5.2.1.

The docs for `import-name` say

>Since version 2.0.9 it is possible to configure this rule with a list of exceptions.

But perhaps they had a bug in version 5.2.1.

Anyway, to test, I installed version tslint-microsoft-contrib@6.2.0 and did `rm -rf node_modules/tslint-config-airbnb/node_modules/tslint-microsoft-contrib`, to force it to resolve to version 6.2.0.

I got some deprecated rules as expected from jumping up a major version, but the `import-name` error went away. So, I think it's an issue with some other config you are extending and maybe not tslint-microsoft-contrib. I will create an issue on tslint-config-airbnb.

---

If you have the above issue, and if you want to hack it (in a reproducible way) to force it to resolve to 6.2.0, you can do the following:
```sh
npm install tslint-microsoft-contrib # get 6.2.0
```

Now you can add `tslint-microsoft-contrib` to `rulesDirectory`, but from what I see, the way tslint works is that it `requires` sub-dependencies before direct dependencies.
In other words, `require('tslint-config-airbnb/node_modules/tslint-microsoft-contrib')` will happen before `require('tslint-microsoft-contrib')`, because `tslint-contrib-airbnb` is inside `extends`, so airbnb's tslint-microsoft-contrib is required first.

You can see it in the output of `tslint --print-config`:
```json
{
  "rulesDirectory": ["tslint-microsoft-contrib"],
  "extends": ["tslint-config-airbnb"]
}
```

becomes

```json
  "rulesDirectory": [
[...]

Issue #876, opened 2019-06-06
Roadmap: tslint-microsoft-contrib -> ESLint


JoshuaKGoldberg:
TSLint is being deprecated soon: see https://github.com/palantir/tslint/issues/4534.

> As you may have read in  [this blog post](https://medium.com/palantir/tslint-in-2019-1a144c2317a9), we plan to deprecate TSLint in 2019 and support the migration to ESLint as the standard linter for both TypeScript & JavaScript. This will not be an immediate deprecation; on the contrary, there is a lot of work to do to ensure a smooth transition to the new tooling without any regressions. 

tslint-microsoft-contrib will follow the timeline in that issue. We'll continue accepting classifications of issues and pull requests, such as new rules, features, and bug fixes, on the same timeline as TSLint. The timeline is:

- __August 1st, 2019__: Stop accepting new _core_ rules. Still accept bug fixes, minor features, and rule enhancements. Custom rules are always an option and can be maintained outside this repo.
- __November 1st, 2019__: Stop accepting features or rule enhancements (with the exception of ones that make migrating to typescript-eslint easier). Still accept bug fixes.
- __January 1st, 2020__: Stop accepting anything except security fixes.
- __December 1st, 2020__: Stop accepting any PRs 🎉

---

Update (August 2019): see **[tslint-to-eslint-config](https://github.com/typescript-eslint/tslint-to-eslint-config)** for a CLI command that migrates TSLint configuration files to ESLint configuration files.

JoshuaKGoldberg:
Copying a couple FAQs from the other issue:

**How can I use existing TSLint rules in ESLint?**

You can use your TSLint rules within ESLint using [typescript-eslint/packages/eslint-plugin-tslint](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin-tslint). It wraps a TSLint configuration and lints your code using TSLint.

**How can I write TypeScript-capable rules in ESLint?**

To write rules in ESLint, see [typescript-eslint/packages/eslint-plugin](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin). That README.md has a list of all their supported rules _(the list has gotten pretty big!)_. The [ROADMAP.md](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md) there has a mapping of existing TSLint rules to the new equivalents.

AndrewCraswell:
[...]

PR #882, opened 2019-07-19
fix: `import-name` options bugs and documentation

dosentmatter:
#### PR checklist

-   [ ] Addresses an existing issue: fixes #0000
    -   [x] Includes tests
-   [x] Documentation update

#### Overview of change:
`import-name` rule has an off-by-one bug where it selects options starting from the wrong index. I updated the tests to match this.

I also noticed that for `camelCase`, it allowed the first letter to be upper-case. `any-case` is either `PascalCase` or `camelCase`, but since `camelCase` allowed the first letter to be upper case, `any-case` and `camelCase` are currently the same. I enforced the first letter to be lowercase.

I also updated the documentation. It's a bit lacking right now. It doesn't explain all of the allowed options.

Not completely related but, I found this bug from #881. In #881, I had other problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,11 +184,16 @@ We recommend you specify exact versions of lint libraries, including `tslint-mic
       </td>
       <td>
         The name of the imported module must match the name of the thing being imported. For special characters (<code>-</code>, <code>.</code>, <code>_</code>) remove them and make the following character uppercase.
-        For example, it is valid to name imported modules the same as the module name: <code>import Service = require('x/y/z/Service')</code> and <code>import Service from 'x/y/z/Service'</code>.
+        For example, it is valid to name imported modules the same as the module name: <code>import Service = require('./x/y/z/Service')</code> and <code>import Service from './x/y/z/Service'</code>.
         But it is invalid to change the name being imported, such as: <code>import MyCoolService = require('x/y/z/Service')</code> and <code>import MyCoolService from 'x/y/z/Service'</code>.
         When containing special characters such as <code>import $$ from 'my-awesome_library';</code> the corresponding configuration would look like <code>'import-name': [true, { 'myAwesomeLibrary': '$$' }]</code>.
         Since version 2.0.9 it is possible to configure this rule with a list of exceptions.
-        For example, to allow <code>underscore</code> to be imported as <code>_</code>, add this configuration: <code>'import-name': [ true, { 'underscore': '_' }]</code>
+        The default is to ignore this rule for external modules <code>ignoreExternalModule: true</code> and to use camelCase <code>case: 'camelCase'</code>.
+        To not ignore this rule for external modules, add this configuration: <code>'import-name': [true, null, null, { ignoreExternalModule: false }]</code>.
+        To allow <code>underscore</code> to be imported as <code>_</code>, add this configuration: <code>'import-name': [true, { 'underscore': '_' }, null, { ignoreExternalModule: false }]</code>.
+        To ignore this rule for <code>react</code> and <code>express</code>, add this configuration: <code>'import-name': [true, null, ['react', 'express'], { ignoreExternalModule: false }]</code>.
+        To use PascalCase: <code>'import-name': [true, null, null, { case: 'PascalCase' }]</code>.
+        To use camelCase or PascalCase: <code>'import-name': [true, null, null, { case: 'any-case' }]</code>.
       </td>
       <td>2.0.5</td>
     </tr>
--- a/package-lock.json
+++ b/package-lock.json
@@ -1299,8 +1299,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -1321,14 +1320,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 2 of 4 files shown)


dosentmatter:
Wonder why Travis CI is failing lint for node 6. The [previous build](https://travis-ci.org/microsoft/tslint-microsoft-contrib/builds/554272966) failed too. [`strict-comparisons`](https://github.com/palantir/tslint/pull/4519) is a new rule. I see a [fix for the false positive](https://github.com/palantir/tslint/pull/4786). Perhaps that will fix it. Haven't looked at why it only fails on node 6 and not other versions.

IllusionMH:
@dosentmatter thank you for the interest and PR. Will check it on weekend.

As for Node 6 - it was shipped with npm v3 that doesn't support `package-lock.json` files. Therefore it installs latest version of TSLint that has false positives instead of fixed version like builds with Node 8+. Error will go away with next TSLint release.

dosentmatter:
Actually, I made a mistake. Currently, `any-case` and `camelCase` are not exactly the same.

The current functionality is the following:
`PascalCase' allows (first letter must be uppercase):
```js
import Foo from './foo';
import Foo from './Foo';
```

`camelCase` allows (first letter must match file name casing):
```js
import foo from './foo';
import Foo from './foo';
```

`any-case' allows (first letter is any case - `PascalCase` | `camelCase`):
```js
import foo from './foo';
import Foo from './foo';
import foo from './Foo';
import Foo from './foo';
```

What this means is that `camelCase` currently enforces that the first letter matches the file name casing. Let me know the desired functionality. If [this wiki on camelCase](https://en.wikipedia.org/wiki/Camel_case) is correct, it says

>Some people and organizations, notably Microsoft,[2](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/x2dbyw72(v=vs.71)) use the term camel case only for lower camel case

README.md
```diff
@@ -184,11 +184,16 @@ We recommend you specify exact versions of lint libraries, including `tslint-mic
...
+        For example, it is valid to name imported modules the same as the module name: <code>import Service = require('./x/y/z/Service')</code> and <code>import Service from './x/y/z/Service'</code>.
         But it is invalid to change the name being imported, such as: <code>import MyCoolService = require('x/y/z/Service')</code> and <code>import MyCoolService from 'x/y/z/Service'</code>.
         When containing special characters such as <code>import $$ from 'my-awesome_library';</code> the corresponding configuration would look like <code>'import-name': [true, { 'myAwesomeLibrary': '$$' }]</code>.
         Since version 2.0.9 it is possible to configure this rule with a list of exceptions.
-        For example, to allow <code>underscore</code> to be imported as <code>_</code>, add this configuration: <code>'import-name': [ true, { 'underscore': '_' }]</code>
+        The default is to ignore this rule for external modules <code>ignoreExternalModule: true</code> and to use camelCase <code>case: 'camelCase'</code>.
+        To not ignore this rule for external modules, add this configuration: <code>'import-name': [true, null, null, { ignoreExternalModule: false }]</code>.
```
IllusionMH:
```suggestion
        To apply this rule for external modules, add this configuration: <code>'import-name': [true, null, null, { ignoreExternalModule: false }]</code>.
```

src/importNameRule.ts
```diff
@@ -233,12 +233,14 @@ function walk(ctx: Lint.WalkContext<Option>) {
     }
 
     function makeCamelCase(input: string): string {
-        return input.replace(
+        let result = `${input.charAt(0)}${input.slice(1)}`;
```
IllusionMH: Doesn't look like this changes anything (without `.toLowerCase()`)

package-lock.json
```diff
@@ -1299,8 +1299,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
```
IllusionMH: Could you please rollback these changes to `package-lock.json`?

src/importNameRule.ts
```diff
@@ -45,15 +45,15 @@ export class Rule extends Lint.Rules.AbstractRule {
 
         if (options.ruleArguments instanceof Array) {
             options.ruleArguments.forEach((opt: unknown, index: number) => {
-                if (index === 1 && isObject(opt)) {
+                if (index === 0 && isObject(opt)) {
```
IllusionMH: This will be a breaking change. I thought that this might be side effect of migration to new rule format, however this behavior was in `6.0.0`.
Looks like better solution is needed to avoid breaking change for those who added one more element to options array to match expected indices

IllusionMH:
I think that it might be possible to avoid breaking change with checks of structure check arguments length and then types arguments e.g.  `options.ruleArguments[0]` is object, `options.ruleArguments[1]` is array, `options.ruleArguments[2]` is object (or null or undefined for elements that were skipped).
If not matched - check with shift by one element (to match current behavior in `master`).

@dosentmatter @JoshuaKGoldberg what do you think on about this approach? Is it possible to do it in general case? Is it worth to implement or better to consider earlier `7.0` with small scope?

dosentmatter:
Hey @IllusionMH, thanks for the review. I'll get to them when I have time. But I just want to respond to one of your comments first before I lose my train of thought.

> I think that it might be possible to avoid breaking change with checks of structure check arguments length and then types arguments e.g. `options.ruleArguments[0]` is object, `options.ruleArguments[1]` is array, `options.ruleArguments[2]` is object (or null or undefined for elements that were skipped).
> If not matched - check with shift by one element (to match current behavior in `master`).
> 
> @dosentmatter @JoshuaKGoldberg what do you think on about this approach? Is it possible to do it in general case? Is it worth to implement or better to consider earlier `7.0` with small scope?

I don't think it is possible to retain complete compatibility with the old code. The old code allowed element 0 to be any type, not just falsy values. It was completely ignored. So let's go over the case with a s length 1 options array:

```js
[{}]
[true]
[undefined]
[null]
[{ 'React': 'react' }]
[{ 'Foo': './foo' }]
```
Although it is unlikely for someone to have an options array like this, all of these used to be ignored. If we now go and pass it to `this.extractReplacements()`, it might break someone's linting. I looked at the code and `ignoreExternalModule` doesn't override the replacements, so `React` and `Foo` will be enforced.

However, maybe we can assume that not many people knew how to skip the first element. The [base tslint.json just used `true`](https://github.com/microsoft/tslint-microsoft-contrib/blob/npm-6.2.0/tslint.json#L123). The docs didn't explain it - are we required to maintain backward compatibility with an undocumented API? All the docs said was that you can pass in replacements to the first element, but it didn't work anyway. Searching up issues, I haven't seen anyone complain until #881.

JoshuaKGoldberg:
Closing this PR per #876. It's always sad to see good work being closed, but we don't have the staffing or relevancy to take on more features. Thanks everyone!

PR #889, opened 2019-10-11
Add missing `dir` to the `no-console`

astorije:
#### PR checklist

-   [ ] Addresses an existing issue
    -   [ ] Includes tests
-   [ ] Documentation update

#### Overview of change:

We noticed this was missing, allowing workarounds. I know this repo is in hiatus now, but even if you do not merge, it could be useful for people stumbling over this wondering why this does not trigger. We are actually going to override the rule by removing the list of methods, disallowing all `console` methods (and that would be my preferred option for the recommended ruleset, even).

I can add tests and docs as needed, if you happen to want to merge this :)

### DIFF
--- a/build-tasks/back-compat/recommended_ruleset.js
+++ b/build-tasks/back-compat/recommended_ruleset.js
@@ -48,7 +48,7 @@ module.exports = {
         'no-backbone-get-set-outside-model': true,
         'no-bitwise': true,
         'no-conditional-assignment': true,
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg:
Ah, thanks for sending this @astorije! You're right that this PR is good as reference, and also right that since we're moving off of TSLint (#876) we won't be merging this PR. 😦

PR #890, opened 2019-10-11
Disallow all console methods

astorije:
#### PR checklist

-   [ ] Addresses an existing issue
    -   [ ] Includes tests
-   [ ] Documentation update

#### Overview of change:

This is an alternative solution for #889, that we prefer, but also more restrictive, so opening both and I'll let you pick.

Just like #889, I can add tests and docs as needed, should you want to merge this.


JoshuaKGoldberg:
(see comments in #889 - thanks for sending)

